PR #3, opened 2022-09-12
Put down starting point for the app.

pentidave:
Added the blueprints for the different views of the app.
Added navigation system via actionBar widget.
Added custom appBar widget.


matteasu:
@gallottino Would you mind reviewing this?

gallottino:
I think that we can merge it. We don't need a deep review for this basic structure.

gallottino (approved):

PR #4, opened 2022-09-14
Remake of the structure using default widgets

pentidave:
Previous version had problems of compatibility, so I remade the structure using the base widgets given by flutter.

- removed widgets folder and corresponding custom widgets.
- added structure.dart file. It contains the Structure widget from which the entire structure of the app is generated.
- added addPage view to visualize when the add button is pressed.
- edited all the pages to include default appBar widget.

Now, the navigation in between pages is managed by the BottomNavigationBar widget.


lib/Pages/structure.dart
```diff
@@ -0,0 +1,91 @@
...
+            BottomNavigationBarItem(
+                icon: Icon(Icons.arrow_back), label: "Movements"),
+            BottomNavigationBarItem(
+                icon: Icon(Icons.bar_chart), label: "Statistics"),
+          ],
+        ));
+  }
+
+  void _onItemTapped(int index) {
+    setState(() {
```
pentidave, line 77: This setState() could be moved inside the if statement.
```
if (index != 2) {
      setState(() {
        _selectedIndex = index;
      });
    } else { ... }
```

PR #20, opened 2022-10-06
Architectural pattern example

AndreBonda:
Clean architecture example. I've organized the source code in a dedicated folder "architecture_example".

Data Layer:
-local database connection.
-data modeling and implementations of repositories.

Domain Layer contains the business logic of the application:
- Domain entities.
- Interfaces of services and repositories.
- Implementations of services that are related to the business logic.

Infrastructure
- Implementations of external services.

Presentation Layer

It will be necessary to add a Dependency Injection implementation to keep all these layers independent.


FedericoBruzzone:
- Would you like to merge this stuff in the main repo?
I'll take a look in your code in the next days 👍

AndreBonda:
It's just an example of an architectural pattern, so you can ignore this commit. If this structure is ok for you, I'll start to apply this pattern to the "original" source code in the next days.

FedericoBruzzone:
Perfect, I think that this PR must be approved by more than one person. 
So, would you start making this arrchitectural pattern right now, or would you start it finished models?

mikev-cw:
Hello! I'm closing this, as we've implemented another pattern meanwhile (not necessarily better...!)
Reach us on Discord if you want to discuss on it! ;)

PR #29, opened 2022-10-20
Custom widget for accounts listing

mikev-cw:
Custom widget to show accounts listing, currently used in dashboard only.

The widget is just return only the graphical elements to build the ui fragment. Dunno if this is how we want to work, but even if want to, we still don't have a definitive logic path for the architecture. 
Widget has an optional image path param usable if we plan to use account logos later, that fallbacks in the default "gear" icon if empty.

To test names and amounts i've created a list of items in home_page.dart that are randomly sent to the Widget for rendering. Notice how long names as "Mediolanum" are splitted in 2 lines. That's seems a bit bad.

Also created lib/customWidgets folder to store our widgets. Is this naming convention good? We need subfolders?


FedericoBruzzone:
Looks good your code.

Some info:
- The folder `custom_widget` is a good choice for me
- Could you show us a picture?
- I think you should use the folder `test/widget` to test your code (*you have to create this folder, I had eliminated it a few days ago beacuse not used*). Maybe we would not touch the `home_page.dart`.

mikev-cw:
- renamed customWidgets folder to custom_widgets
- removed random test accounts from home_page.dart
- added unit test for this widget
- check pics attached: with and without account logo
Screenshot_20221025_110637
Screenshot_20221025_000506

FedericoBruzzone:
Looks good @mikev-cw 
For your first commit that's fine, but for the next few times it would be better not to do so many useless commits.

FedericoBruzzone:
image

Tests have not been launched
@mikev-cw @matteasu

PR #36, opened 2022-11-28
Design: implement Transactions page

GBergatto:
As I said on the [Discord](http://mr.rip/discord) channel, I'm working on the "Transactions" page.

I've implemented the page structure with a collapsible header and 3 tabs. I've added a placeholder `ListView` in each tab.

Improvements for the header:
- [x] Disable all buttons during the transition
- [x] Make the header [floating](https://api.flutter.dev/flutter/material/SliverAppBar/floating.html)
- [x] Make the header snap to fully extended/collapsed when it's almost there

In addition, I've edited `ThemeData` and defined some colors based on the design doc on Figma.


mikev-cw:
Please use only english in code fragments and filenames. 
Also change "movements" to "transactions", just to keep consistency on how we refer to that page

lib/pages/transactions_page/transactions_page.dart
```diff
@@ -0,0 +1,63 @@
...
+
+class TransactionsPage extends StatefulWidget {
+  @override
+  _TransactionsPageState createState() => _TransactionsPageState();
+}
+
+class _TransactionsPageState extends State<TransactionsPage>
+    with SingleTickerProviderStateMixin {
+  static const List<Tab> myTabs = <Tab>[
+    Tab(text: "Elenco", height: 35),
```
FedericoBruzzone, line 16: Can you use the english language?

lib/pages/transactions_page/widgets/custom_sliver_delegate.dart
```diff
@@ -0,0 +1,169 @@
...
+          "Settembre 2022",
+          style:
+              Theme.of(context).textTheme.displayMedium?.copyWith(fontSize: 14),
+        ),
+        Text(
+          "$amount €",
+          style: TextStyle(
+              color: (amount > 0) ? Colors.green : Colors.red,
+              fontWeight: FontWeight.bold,
+              fontFamily: 'SF Pro Text'),
```
FedericoBruzzone, line 155: I'd suggest you to use a macro for the font name (and replace it where necessary, also in the other widgets)

I apologize if you do not know what a macro is, I mean something like a global variable.

Let's talk about whether it is better to create a new file for the macros.

(https://docs.flutter.dev/cookbook/design/fonts)

lib/pages/transactions_page/widgets/month_selector.dart
```diff
@@ -0,0 +1,80 @@
...
+                Text(
+                  "Settembre 2022",
+                  style: Theme.of(context).textTheme.displayMedium,
+                ),
+                Text(
+                  "$amount €",
+                  style: TextStyle(
+                      color: (amount > 0) ? Colors.green : Colors.red,
+                      fontWeight: FontWeight.bold,
+                      fontFamily: 'SF Pro Text'),
```
FedericoBruzzone, line 56: Same this above

FedericoBruzzone (changes requested):

lib/pages/transactions_page/transactions_page.dart
FedericoBruzzone: Exactly, maybe a good choice would be to write in English.
What do you think about it?

GBergatto: IMO we should build the app in English as the default language to make it more international. Later on, we can add support for different languages including Italian. However, we should probably discuss this on Discord as a project-wide change.

lib/pages/transactions_page/widgets/custom_sliver_delegate.dart
GBergatto: I wasn't sure how we planned to handle fonts and colors across the app, so I temporarily hardcoded them.
One way would be to have a file where we define global variables (macros), but we could also define a Theme and make all widgets inherit from it.
I can use a macro but I don't know where to put it

lib/pages/transactions_page/transactions_page.dart
FedericoBruzzone: Yes, probably we should discuss about it on Discord. But IMHO it is the best choice. :D

lib/pages/transactions_page/widgets/custom_sliver_delegate.dart
FedericoBruzzone: I think that at this project state, we could use a file and in the future make all widgets inherit from the theme. What do you think?

mikev-cw: Aren't fonts already defined here?
https://github.com/RIP-Comm/sossoldi/blob/20f13797f5bf5d578e66525892245020c2f00ccc/lib/main.dart#L35

lib/pages/transactions_page/transactions_page.dart
GBergatto: I changed all titles to English

lib/pages/transactions_page/widgets/custom_sliver_delegate.dart
GBergatto: Done

FedericoBruzzone:
Can we merge this PR?

GBergatto:
As far as I'm concerned this PR is ready to be merged. I'll explain its contents here.

I've built the structure of the "Transactions" page and created the collapsible header used to navigate between the 3 tabs. I've added a scrollable list to each tab as a placeholder.

I've also edited the `TextTheme` and defined some colors as constants in `constants/style.dart` based on the design doc on [Figma](https://www.figma.com/file/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?node-id=1736%3A4229&t=aW9iVSQgLMA7lQWR-4). This should address issue [#40](https://github.com/RIP-Comm/sossoldi/issues/40). 

The main colors and font styles are included in `ThemeData` and should be accessed with `Theme.of(context)`. The intended use of `constants/style.dart` is to be imported every time colors and styles (or other constants) not included in `ThemeData` are needed for a certain widget. 
I kept `ThemeData` inside `main.dart` instead of moving it to `style.dart` because I don't think it should be imported along with the constants but rather inherited via `Theme.of(context)`.

Many elements of the app now look weird as they are inheriting from the wrong property of `TextTheme`. Also, most widgets have a hard-coded color that overrides the new theme, so its effect is mostly hidden.

There were still 2 undone tasks in this PR but I decided to remove them as they are more complex than originally anticipated and not crucial for this early stage. The animation of the header will need some improvements in the future, but the function is there.

Finally, I'd like to make the following two suggestions:
1. Using relative imports for our files. See [#42](https://github.com/RIP-Comm/sossoldi/issues/42)
2. Breaking down pages into separate widgets. See [#43](https://github.com/RIP-Comm/sossoldi/issues/43)

PR #37, opened 2022-12-04
Created line chart widget + Mock examples in Homepage and Account Modal

theperu:
Created the upper part of the Homepage based on the Figma (https://www.figma.com/file/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?node-id=870%3A1075&t=qhyvRfo8U3EACCrK-4) using fl_chart. One thing that is not currently implemented is the information showed when the user taps on one day of the chart (curretly shows the raw data)
There are also some changes that VSCode did automatically in order to let me use the debugger on macOS and the iOS emulator.


mikev-cw:
Thank you!
I think you should remove all committed files inside ios and macos folders from this PR, as they're somehow related to the single system running the build, i think. I'm unable to checkout your main branch because of this (my podfiles are .gitignored).

If you need it, see [here](https://ffeathers.wordpress.com/2020/03/21/removing-an-updated-file-from-a-pr-on-github/) to find out how to do that.

I general we don't want to merge software automatic edited files. Afterwards we can check your work. Thank you!

theperu:
Sorry about that, somehow I thought they were just default files for all. I reset the folders for iOS and macOS, hopefully now it fine

theperu:
I tried to implement what we discussed on Discord before. Hopefully I understood what you asked, now the placeholder data is not in the widget file but is written on the home_page.dart

theperu:
In the next days I will also try to add the test and I will also talk to Federico P. about the behaviour of the chart when the current and the previous month differs in the number of days.

FedericoBruzzone:
Need rebase :D

theperu:
A little update on my end. A made the chart class generic so that we can reuse it and created a modal just to verify that it works. I didn't commit it because I still need to do a little bit of clean-up and add some comments so that who will use it later will have an easier path

mikev-cw:
Ok, great!!
Maybe you can rebase your fork to get last changes to the theme management (see #36) and use this ultimate references in your work.
Also check #43 for the correct approach to use when importing libs, and #42 to make sure you're using the correct paths for widgets

theperu:
Ok, I tried to make the corrections and hopefully everything is fine now. The linechart widget has some parameters in order to display two lines, those are: the data point lists, the colors, maxY, minY, the # of days and the background color. If we need to display just one line we can pass an empty list. 
To make a little more clear the use I also implemented the modal that opens when a user taps on one of the accounts from the dashboard.
I also tried to implement one test but at the moment is not really useful, if someone has an idea on what I should do and has some resources I'll try to make something better.

FedericoBruzzone (approved):

mikev-cw:
Ok cool, very good work!
Can you just remove "macos" folder and "pubspec.lock" file from the PR so I can checkout? Thanks!

theperu:
@mikev-cw Now it should be fine. I always try to exclude them but sometimes they still go through

FedericoBruzzone (approved):

Issue #44, opened 2022-12-20
Internal alerting system


mikev-cw:
We need a class that handles displayable alerts to the user. 
I mean messages like "ok! Your transactions has been added" or error messages like "a category is needed to save the transaction"

NandayDev:
I can do this. Is there UI design to follow, or should I just create a generic alert dialog?

mikev-cw:
Good! I don't think there is a render on Figma about that. 
You can just try to do a generic dialog, with classic success/warning/error/info design.
Thank you!

mikev-cw:
Cool! 
Just fork this repo, do your commits, then open a PR. We'll merge from there.
Thank you!

NandayDev:
There you go https://github.com/RIP-Comm/sossoldi/pull/46

Issue #43, opened 2022-12-20
Use relative path to import files


GBergatto:
To keep a consistent coding style, I suggest we use relative paths when importing files from within our app.

```
// Before
import 'package:sossoldi/model/example.dart';

// After
import 'model/example.dart';
```

As long as we use it consistently, any of the two styles will have the same performance. However, it's important not to mix them as that might lead to some issues.

The official guide on dart recommends using [relative import paths](https://dart.dev/guides/language/effective-dart/usage#prefer-relative-import-paths).
See also https://stackoverflow.com/q/59693195/14157994 for pros and cons.

The reason why I prefer relative paths is that they are shorter and help differentiate between our files (within `lib`) and external packages.

mikev-cw:
Now all paths are correctly switched to relative. 
Thank you, I'm closing this.

PR #46, opened 2023-01-04
#44 Added alert dialog class

NandayDev:
To show a dialog, you can either use these straightforward functions to simply show some text with an "OK" button to dismiss it:
- void showInfoDialog(BuildContext context, String text)
- void showSuccessDialog(BuildContext context, String text)
- void showWarningDialog(BuildContext context, String text)
- void showErrorDialog(BuildContext context, String text)

or instantiate an AlertDialogBuilder, with different parameters, and then invoke show(). I added a comment on each property (lines 18-52), let me know should you have any question

### DIFF
--- a/lib/custom_widgets/alert_dialog.dart
+++ b/lib/custom_widgets/alert_dialog.dart
@@ -71,6 +71,12 @@ void showInfoDialog(BuildContext context, String text) =>
     AlertDialogBuilder(text: text, dialogType: AlertDialogType.info, primaryActionText: "OK").show(context);
 
 ///
+/// Shows a success dialog with given text
+///
+void showSuccessDialog(BuildContext context, String text) =>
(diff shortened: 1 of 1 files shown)


mikev-cw:
Can you show us how can we show a dialog, after importing the class?
Thank you

lib/custom_widgets/alert_dialog.dart
```diff
@@ -0,0 +1,147 @@
+import 'package:flutter/material.dart';
+import 'package:sossoldi/main.dart';
```
mikev-cw, line 2: Use relative path to import internal files, as per #43

NandayDev: Ok done with the latest commits

NandayDev:
> Can you show us how can we display a dialog, after importing the class?
Thank you

@mikev-cw Top comment updated

PR #58, opened 2023-01-14
settings_page

AntonioMiclaus:
I've added the settings_page with the four main options.

### DIFF
--- a/macos/Flutter/GeneratedPluginRegistrant.swift
+++ b/macos/Flutter/GeneratedPluginRegistrant.swift
@@ -1,14 +1,15 @@
 //
 //  Generated file. Do not edit.
 //
-
 import FlutterMacOS
 import Foundation
 
+import path_provider_macos
 import sqflite
 import sqlite3_flutter_libs
 
 func RegisterGeneratedPlugins(registry: FlutterPluginRegistry) {
+  PathProviderPlugin.register(with: registry.registrar(forPlugin: "PathProviderPlugin"))
   SqflitePlugin.register(with: registry.registrar(forPlugin: "SqflitePlugin"))
   Sqlite3FlutterLibsPlugin.register(with: registry.registrar(forPlugin: "Sqlite3FlutterLibsPlugin"))
 }
--- a/macos/Flutter/GeneratedPluginRegistrant.swift
+++ b/macos/Flutter/GeneratedPluginRegistrant.swift
@@ -5,12 +5,10 @@
 import FlutterMacOS
 import Foundation
 
-import path_provider_macos
 import sqflite
 import sqlite3_flutter_libs
 
 func RegisterGeneratedPlugins(registry: FlutterPluginRegistry) {
-  PathProviderPlugin.register(with: registry.registrar(forPlugin: "PathProviderPlugin"))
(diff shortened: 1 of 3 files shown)


theperu:
Great job! I tried this on both iOS & Android and the only thing that looks a bit strange to me is the upper part of the screen for two reasons:
- Having a grey background seems strange and it is different from the other parts of the app, maybe I would change it to the same white used for the rest of the page. If the icon and the text are not easily visible we could set them to the same blue of the other ones.
- "Back" is not right besides the icon (on iOS is shown in the middle in fact). If you struggle on moving it to the left a work around could be to remove the word "Back" leaving only the icon and having settings written in the same line in center instead of where is now

FedericoBruzzone:
Can you send a picture? I can't test your code.

theperu:
Here is how it looks on iOS & Android
Screenshot 2023-01-15 at 10 30 05
Screenshot 2023-01-15 at 10 31 18

FedericoBruzzone:
I agree with what you said. 👍

AntonioMiclaus:
I've updated the appbar by removing the "back" text  and the grey background and added the text "Settings" in the center with also the dark blue back arrow.

theperu:
Hi! Everything looks fine to me now. Just a little question, I saw that in your commits you also pushed changes to macos/Flutter/GeneratedPluginRegistrant.swift and pubspec.lock but it don't understand why. Can you help me with that?

AntonioMiclaus:
I noticed these files changed automatically when I tested the code with VScode + android studio. The added parts may be requirements. The parts removed may not be needed anymore or maybe only for the device I tested on. The only file I change manually is the settings_page.dart

theperu:
Understood, I have the feeling that a PR like this should contain only the changes on settings_page.dart. Maybe @FedericoBruzzone or @mikev-cw can take a look since I am not that familiar with code reviews

macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -5,12 +5,10 @@
 import FlutterMacOS
 import Foundation
 
-import path_provider_macos
 import sqflite
 import sqlite3_flutter_libs
 
 func RegisterGeneratedPlugins(registry: FlutterPluginRegistry) {
-  PathProviderPlugin.register(with: registry.registrar(forPlugin: "PathProviderPlugin"))
```
FedericoBruzzone, line 13: In my opinion, this part shouldn't be in this PR.
I suggest you to add the missing code and use `--amend` command to update this PR :D

FedericoBruzzone (changes requested):

macos/Flutter/GeneratedPluginRegistrant.swift
AntonioMiclaus: Thanks, I've restored the original file.

mikev-cw:
@theperu Yeah! Agreed.
@AntonioMiclaus seems that files are still in the PR. Maybe you want to open a new one?

FedericoBruzzone:
To remove useless commit use: 

git rebase -i HEAD~5 

And type "drop" in place of "pick".

AntonioMiclaus:
Thanks for all the comments and help. I'm gonna close this request and open a new one in order to avoid further confusion.

PR #62, opened 2023-01-19
Riverpod State Management Integrated + Transaction page (in progress)

lucaantonelli:
1. I have integrated riverpod in the app, you can check the new providers folder, for now:
   - it works in the dashboard page for the accounts and transactions sections
   - also when choosing the category in transaction
   - for other small things like page index in the navigation bar

2. Work in progress for the transactions page (it works, but you can only set the amount, and the category that is mandatory for now), if you add a transaction it will be visible in the home.
3. Bottom Navigation Bar styling
4. Other minor overall improvements


lucaantonelli:
I will do a new PR as i made some mistakes with commit :)

PR #64, opened 2023-01-21
Add Splash Screen Icon for Web, Android and iOS

Filippoml:
The following PR implements a splash screen icon for Web, Android and iOS. Unfortunately, I couldn't find a way to do this for Windows as well.

The splash screens have been generated through the plugin [flutter_native_splash](https://pub.dev/packages/flutter_native_splash), which has been added to the project in case it may be needed in the future to update the icon.

PLEASE NOTE: The splash screen may not appear when you launch the app from Android Studio. However, it should appear when you launch by clicking on the launch icon in Android. 

PLEASE NOTE: There are a number of reports that non-Google launchers do not display the launch image correctly. If the launch image does not display correctly, please try the Google launcher to confirm that this package is working.

PLEASE NOTE: The splash screen does not appear when you launch the app from a notification. Apparently, this is the intended behavior on Android 11.

image


FedericoBruzzone (approved):

theperu:
Nice work! I tried this and I noticed that on iOS the image is a lot smaller compared to Android. Although this is not a big issue you could look in to that and see if there is an easy fix (if there isn't one it's fine we will keep it as is). You can see the video down below for reference.

https://user-images.githubusercontent.com/24766687/213923854-42b48d4a-d2c0-4201-862c-5b7b175b7169.mov

FedericoBruzzone:
Yes, I think it's too small

Filippoml:
I see... unfortunately I am on Windows so I would have to find a way to run a iOS emulator.
The reason why is a bit small is because on Android I don't have a lot of space available for the icon.
One thing that we could do is to use the logo without the text "sossoldi", in this way it will appear bigger than now.

theperu:
From my understanding we can use different images for Android and iOS so since on the current image there is a lot of empty space we could make the logo bigger just for the iOS images and that could solve the problem

Filippoml:
The problem is not with iOS, I have noticed that the splash screen icon looks smaller on displays with higher resolution, I am going to see if I can find a solution for that.

theperu:
Got it, I would still suggest the same solution. The images that you used have a lot of empty space and that is why it is showing like this in some devices, it should be pretty easy to modify them in order to achieve that. If you have any problem doing that you can ask on Discord.

Filippoml:
I have cropped the splash screen icons, can anybody test on iOS?

theperu:
Hi! Tested it again on iOS and they still look the same, I think it is because by cropping the images we lost the original resolution. I am sorry if I wasn't clear with my suggestion last time but what we need I think are the images with the resolution that we had at the start but without the white space. So the images should look like the ones that you cropped but with the starting resolution. If you have any kind of trouble with this I suggest you to reach out on Discord channel for the frontend (which is "Graphical User Interface").

https://user-images.githubusercontent.com/24766687/217645898-5cee5ad6-3cfc-471f-b179-54e587058804.mov

FedericoBruzzone (approved):

theperu:
Although there are a couple of things that we should sort out on this PR I think we can easily fix them in the future. So I am ok with merging this one

PR #66, opened 2023-01-25
Implementation of tap to close keyboard on the transaction creation screen

marianialessandro:
A GestureDetector was used to implement this feature


mikev-cw:
Can you remove all files into ios and macos folders? Also pubspec.lock.
Ideally this PR should only have add_page.dart

marianialessandro:
> Can you remove all files into ios and macos folders? Also pubspec.lock.
> Ideally this PR should only have add_page.dart

Sure

PR #74, opened 2023-02-18
DB correction + minor graphic fix

lucaantonelli:
Resolved some error with the new db fields.
Removed boolean type as it doesn't exist in the sqlite plugin: `bool is not a supported SQLite type. Use INTEGER and 0 and 1 values.` ([source](https://pub.dev/packages/sqflite#supported-sqlite-types))

### DIFF
--- a/lib/custom_widgets/accounts_sum.dart
+++ b/lib/custom_widgets/accounts_sum.dart
@@ -21,7 +21,7 @@ class AccountsSum extends StatelessWidget with Functions {
   Widget build(BuildContext context) {
     return Container(
       width: 160.0,
-      margin: const EdgeInsets.fromLTRB(0, 4, 16, 4),
+      margin: const EdgeInsets.fromLTRB(0, 4, 16, 6),
       decoration: BoxDecoration(
         color: const Color(0xFFD7E2ED),
         borderRadius: BorderRadius.circular(8),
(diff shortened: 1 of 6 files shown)


mikev-cw:
I'm getting this on Mac OS:

> I/flutter (14356): *** WARNING ***
I/flutter (14356): 
I/flutter (14356): Invalid argument true with type bool.
I/flutter (14356): Only num, String and Uint8List are supported. See https://github.com/tekartik/sqflite/blob/master/sqflite/doc/supported_types.md for details
I/flutter (14356): 
I/flutter (14356): This will throw an exception in the future. For now it is displayed once per type.

Is your commit related?

lucaantonelli:
> I'm getting this on Mac OS:
> 
> > I/flutter (14356): *** WARNING ***
> I/flutter (14356): 
> I/flutter (14356): Invalid argument true with type bool.
> I/flutter (14356): Only num, String and Uint8List are supported. See https://github.com/tekartik/sqflite/blob/master/sqflite/doc/supported_types.md for details
> I/flutter (14356): 
> I/flutter (14356): This will throw an exception in the future. For now it is displayed once per type.
> 
> Is your commit related?
> 
> EDIT: I'm pretty sure this is an old Warn, so the correct question should be: is your commit intended to solve also this?

Yes, the commit solves also that problem, i think it shows up to you for the cached db, if you remove and reinstall the app it shouldn't appear again, at least for me that worked fine.

PR #75, opened 2023-02-21
Add active to budget table

sainzrow:
Added active column as integer with 0 and 1 values.
Added new selectAllActive query to pick up just the active budgets.
Replaced physical delete with logical one.

### DIFF
--- a/lib/model/budget.dart
+++ b/lib/model/budget.dart
@@ -9,6 +9,7 @@ class BudgetFields extends BaseEntityFields {
   static String name = 'name';
   static String idCategory = 'idCategory'; // FK
   static String amountLimit = 'amountLimit';
+  static String active = 'active';
   static String createdAt = BaseEntityFields.getCreatedAt;
   static String updatedAt = BaseEntityFields.getUpdatedAt;
 
@@ -25,12 +26,14 @@ class BudgetFields extends BaseEntityFields {
 class Budget extends BaseEntity {
   final int idCategory;
   final num amountLimit;
+  final int active;
   final String? name;
 
   const Budget(
       {int? id,
       required this.idCategory,
       required this.amountLimit,
+      required this.active,
       String? this.name,
       DateTime? createdAt,
       DateTime? updatedAt})
@@ -40,12 +43,14 @@ class Budget extends BaseEntity {
           {int? id,
           int? idCategory,
           num? amountLimit,
+          int? active,
           DateTime? createdAt,
           DateTime? updatedAt}) =>
       Budget(
           id: id ?? this.id,
           idCategory: idCategory ?? this.idCategory,
           amountLimit: amountLimit ?? this.amountLimit,
+          active: active ?? this.active,
           createdAt: createdAt ?? this.createdAt,
           updatedAt: updatedAt ?? this.updatedAt);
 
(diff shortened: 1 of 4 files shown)


mikev-cw:
Cool!
For consistency with the project, I think we should use same approach used -for instance- in the `mainAccount` field of `bankAccount` model: we use `bool` true/false in dart object, then the model takes care of converting `bool` to `int` 1/0 on functions that read/write on db.
Can you adjust your work like this? Thank you!

sainzrow:
Sure, i'll update the PR.
However, i think there might be a problem with the insert/update of the bankAccount model because of the following row:
**Map<String, Object?> toJson({bool update = false}) => {
        ...
BankAccountFields.mainAccount: mainAccount,**
        ...
      };**
as it seems to put a boolean type into an int. 

Should i fix this in my budget model?
Thanks

lib/model/budget.dart
```diff
@@ -9,60 +9,73 @@ class BudgetFields extends BaseEntityFields {
...
   final num amountLimit;
-  final String? name;
+  final bool active;
 
   const Budget(
       {int? id,
       required this.idCategory,
+      required this.name,
       required this.amountLimit,
-      String? this.name,
```
mikev-cw, line 34: Why you added budget name as required field in this object? You also added this field in create table statement. It was removed on #71

lib/model/transaction.dart
```diff
@@ -130,8 +130,8 @@ class Transaction extends BaseEntity {
       recurring: json[TransactionFields.recurring] == 1 ? true : false,
       recurrencyType: json[TransactionFields.recurrencyType] as String?,
       recurrencyPayDay: json[TransactionFields.recurrencyPayDay] as int?,
-      recurrencyFrom: json[TransactionFields.recurrencyFrom] != null ? DateTime.parse (TransactionFields.recurrencyFrom) : null,
-      recurrencyTo: json[TransactionFields.recurrencyTo] != null ? DateTime.parse (TransactionFields.recurrencyTo) : null,
+      recurrencyFrom: json[TransactionFields.recurrencyFrom] != null ? DateTime.parse(TransactionFields.recurrencyFrom) : null,
+      recurrencyTo: json[TransactionFields.recurrencyTo] != null ? DateTime.parse(TransactionFields.recurrencyTo) : null,
```
mikev-cw, line 134: This blank space was already been removed on #74, seems you're working on old files

lib/model/budget.dart
```diff
@@ -104,20 +125,17 @@ class BudgetMethods extends SossoldiDatabase {
...
   Future<int> deleteById(int id) async {
     final database = await SossoldiDatabase.instance.database;
 
-    return await database.delete(budgetTable,
-        where:
-        '${BudgetFields.id} = ?',
-        whereArgs: [id]);
+    // logical delete by setting active to 0
+    return await database
+        .rawUpdate('UPDATE Budget SET active = 0 WHERE id = ?', [id]);
```
mikev-cw, line 139: Not sure if we want this behavior here. I think this delete function should actually delete the record on db, and when the user removes a budget, our BE logic should use `updateItem()` function to change the `active` field to 0.

mikev-cw (commented):
You've commit some changes reverted from previous PR, like all minor graphical changes already done in #74. Also can you go deeper on the `name` field thing?

Thank you!

lucaantonelli:
> Sure, i'll update the PR. However, i think there might be a problem with the insert/update of the bankAccount model because of the following row: **Map<String, Object?> toJson({bool update = false}) => { ... BankAccountFields.mainAccount: mainAccount, ... };** as it seems to put a boolean type into an int.
> 
> Should i fix this in my budget model? Thanks

You're right, it is needed to set `BankAccountFields.mainAccount: mainAccount ? 1 : 0`, i forgot it in the last commit. It is also needed in currency.dart for mainCurrency and in transaction.dart for recurring, if you can do that so is all in just one PR.

FedericoBruzzone (approved):

mikev-cw:
I'm merging this as discussed on Discord: we'll maybe fix reviewed issues later

PR #80, opened 2023-03-21
Transactions working implementation + minor fix

lucaantonelli:
Now transactions works fine, insert/update/delete and income/expense/transfer.
The only missing implementation is for recurring payment.

Known Issue: A glitch when disposing the transaction modal on delete

### DIFF
--- a/lib/constants/functions.dart
+++ b/lib/constants/functions.dart
@@ -4,7 +4,8 @@ import '../constants/style.dart';
 import '../model/transaction.dart';
 
 mixin Functions {
-  String numToCurrency(num value) {
+  String numToCurrency(num? value) {
+    if(value == null) return '';
     return value.toStringAsFixed(2).replaceAll(".", ",");
   }
 
--- a/lib/custom_widgets/transactions_list.dart
+++ b/lib/custom_widgets/transactions_list.dart
@@ -2,6 +2,8 @@ import 'dart:ui';
 
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import '../pages/add_page/add_page.dart';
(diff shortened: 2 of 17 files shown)


theperu:
While checking this I noticed a couple of things:
1. It looks like we don't have any categories from the add transaction page which to me it doesn't seem right
2. When I try to add a transfer between two accounts and I come back to the dashboard I see an error
Not sure if this is specifically related to this PR but we should look into that.
Also if I close and reopen the app the error for the account transfer goes away kinda like it wasn't saved. Am I doing something wrong? 
I uploaded a video of my tests [here](https://drive.google.com/file/d/17gInj70WTZjmdFKu67atihm1V5G2P47q/view?usp=share_link)

lucaantonelli:
> While checking this I noticed a couple of things:
> 
> 1. It looks like we don't have any categories from the add transaction page which to me it doesn't seem right
> 2. When I try to add a transfer between two accounts and I come back to the dashboard I see an error
>    Not sure if this is specifically related to this PR but we should look into that.
>    Also if I close and reopen the app the error for the account transfer goes away kinda like it wasn't saved. Am I doing something wrong?
>    I uploaded a video of my tests [here](https://drive.google.com/file/d/17gInj70WTZjmdFKu67atihm1V5G2P47q/view?usp=share_link)

Have you tried removing app data or delete the app and reinstall? It seems you could have an old version of the "fake data" in the db, otherwise if it continues to not work can you share the console log?

theperu:
You are right, it looks like something went wrong with the iOS emulator and I didn't notice. I tried again and it looks fine now. Great work! To me it looks like we can merge this if you are not planning to implement the recurring payment right away

mikev-cw:
I confirm it's working on Android too. Super cool, let's merge!

PR #84, opened 2023-04-09
Improvements to "Transactions" page

GBergatto:
I improved the List and Categories tabs. Here are two screenshots of how they look now.
<p float="left">
 
   
</p>

Currently, all transactions are queried from the database regardless of their date. We should query _only the ones in the selected time frame_. Also, the totals for the day and for each category are calculated in the build method of the widgets, which is not ideal. It might be better to retrieve them directly from the database with a designated query. The same goes for the groupings of transactions in their categories in the "Categories" tab.

The list of expandable categories in the "Categories" tab is wrapped in a `Sizebox` whose size is calculated every time a category is expanded. While a category gets collapsed, a `RenderFlex overflowed` is sometimes thrown, because the new size is computed before the end of the animation. After the animation is over, everything is fine.
If you can think of a better solution, please tell me.

## What's missing
- bank account and category of the transaction in the "Categories" tab are shown as numbers instead of names
- colors of the categories for icons and pie chart (still missing from database)
- more specific queries and more efficient calculation of totals

### DIFF
--- a/lib/pages/transactions_page/transactions_page.dart
+++ b/lib/pages/transactions_page/transactions_page.dart
@@ -6,8 +6,10 @@ import 'widgets/accounts_tab.dart';
 import 'widgets/list_tab.dart';
 
 class TransactionsPage extends StatefulWidget {
+  const TransactionsPage({super.key});
+
   @override
-  _TransactionsPageState createState() => _TransactionsPageState();
+  State<TransactionsPage> createState() => _TransactionsPageState();
 }
 
 class _TransactionsPageState extends State<TransactionsPage>
--- a/lib/pages/transactions_page/widgets/categories_pie_chart.dart
+++ b/lib/pages/transactions_page/widgets/categories_pie_chart.dart
@@ -1,18 +1,23 @@
 import "package:flutter/material.dart";
 import 'package:fl_chart/fl_chart.dart';
-import 'package:sossoldi/constants/style.dart';
 
+import '../../../constants/functions.dart';
+import '../../../constants/style.dart';
 import '../../../model/category_transaction.dart';
 
-class CategoriesPieChart extends StatelessWidget {
+class CategoriesPieChart extends StatelessWidget with Functions {
   const CategoriesPieChart({
     required this.notifier,
     required this.categories,
+    required this.amounts,
+    required this.total,
     Key? key,
   }) : super(key: key);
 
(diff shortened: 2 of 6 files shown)


FedericoBruzzone (changes requested):
Thanks for this PR, in addition to the error already shown on discord, I noticed that some comments are written with lowercase letter. In the moment when you'll update the PR, please change this too.

GBergatto:
> Thanks for this PR, in addition to the error already shown on discord, I noticed that some comments are written with lowercase letter. In the moment when you'll update the PR, please change this too.

The error shown on discord is caused by a typing error at line 149 of the file `lib/database/sossoldi_database.dart`
```sql
(14, "Leisure", "subscriptions", '', null, '${DateTime.now()}', '${DateTime.now()}');
15, "Salary", "work", '', null, '${DateTime.now()}', '${DateTime.now()}');
```
The semicolon at the end of the line should be a comma.

I have fixed this on my local fork but haven't committed the change as I thought it would have been out of scope for this PR. @lucaantonelli said he was fixing a few issues with the database.

If you want me to, I can make all comments start with uppercase letters. However, I think it would be a bit overkill, considering that, as far as I know, no convention about comments has been established. (Maybe we should establish one now)

mikev-cw (approved):

mikev-cw:
> > Thanks for this PR, in addition to the error already shown on discord, I noticed that some comments are written with lowercase letter. In the moment when you'll update the PR, please change this too.
> 
> The error shown on discord is caused by a typing error at line 149 of the file `lib/database/sossoldi_database.dart`
> 
> ```sql
> (14, "Leisure", "subscriptions", '', null, '${DateTime.now()}', '${DateTime.now()}');
> 15, "Salary", "work", '', null, '${DateTime.now()}', '${DateTime.now()}');
> ```
> 
> The semicolon at the end of the line should be a comma.
> 
> I have fixed this on my local fork but haven't committed the change as I thought it would have been out of scope for this PR. @lucaantonelli said he was fixing a few issues with the database.
> 
> If you want me to, I can make all comments start with uppercase letters. However, I think it would be a bit overkill, considering that, as far as I know, no convention about comments has been established. (Maybe we should establish one now)

Agreed.
Just fixed the typo with a [direct commit](https://github.com/RIP-Comm/sossoldi/commit/79fdd250967c7dbb316a461683a25e071e321ff7).

PR #85, opened 2023-04-28
Updated accounts and categories management

lucaantonelli:
- Added logo
- The splash screen now should be entirely visible and not cropped by a white circle (it is an android 12 and newer behaviour, so it looks a bit small on the latest android versions)
- Added accounts and categories management (insert/update/delete) with new styling
- Improved homepage
- Settings page with new layout

Known Issues:
- If you delete the main account or deselect it, it throws an error because as of now, the main account is mandatory. We should decide if it can stay as is, maybe automatically setting the main account as the most used (in last year or something like that).


theperu:
Great job! I tried the app a little bit and noticed a few things:
- For some reason the app icon is a bit weird when you look all the backgrounded apps 
Screenshot 2023-04-29 at 12 58 00
- I am able to submit a transfer without the second account selected, when I comeback to the home then I see an error in the last transactions section. 
https://user-images.githubusercontent.com/24766687/235299507-b55753ff-c0eb-439d-b74d-02c447a58f3a.mov
- This is not related to your PR but is something that I noticed. Although I can create new transactions the values for the accounts don't get updated.
- The categories for income and expenses should be separated (ex. this means that when I am submitting a new expense I shouldn't see "income")

GBergatto:
I noticed that `color` is an integers used as the index for a list of default colors. Wouldn't it be better to save the HEX code instead and convert it directly to the `Color` class? With the current solution, if we were to edit the list of colors, we'd have to change all indexes across the database.

mikev-cw:
Super cool! I'm merging this. As discussed on Discord, we will take care of issues raised by @theperu and @GBergatto on other PRs.

PR #93, opened 2023-05-12
Added More Info & Privacy Policy pages

theperu:
As we discussed yesterday I tried to write some Privacy Policy for our app, of course I am not a lawyer and I don't know if this are fine or not. I tried to be safe and write sections that should also cover future changes (see Service Providers, Log Data and Information Collection and Use). 
I also add the version of the app and collaborators (this page is not ready yet but we will update it in the future)

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -26,7 +26,7 @@ apply plugin: 'kotlin-android'
 apply from: "$flutterRoot/packages/flutter_tools/gradle/flutter.gradle"
 
 android {
-    compileSdkVersion flutter.compileSdkVersion
+    compileSdkVersion 33
     ndkVersion flutter.ndkVersion
 
     compileOptions {
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlin_version = '1.6.10'
+    ext.kotlin_version = '1.8.20'
     repositories {
         google()
         mavenCentral()
--- a/lib/custom_widgets/account_modal.dart
+++ b/lib/custom_widgets/account_modal.dart
@@ -2,17 +2,30 @@ import 'dart:ui';
 
 import 'package:fl_chart/fl_chart.dart';
 import 'package:flutter/material.dart';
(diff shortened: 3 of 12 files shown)


FedericoBruzzone (approved):
LGTM👍

theperu:
Not sure on why this is failing now since there isn't much new with the latest commit. It looks like it is SDK related and we also had a warning before, I would leave it to you to decide what is best to do

FedericoBruzzone:
> Not sure on why this is failing now since there isn't much new with the latest commit. It looks like it is SDK related and we also had a warning before, I would leave it to you to decide what is best to do

```
Dependency 'androidx.core:core:1.10.1' requires 'compileSdkVersion' to be set to 33 or higher.
     Compilation target for module ':app' is 'android-31'

Dependency 'androidx.annotation:annotation-experimental:1.3.0' requires 'compileSdkVersion' to be set to 33 or higher.
     Compilation target for module ':app' is 'android-31'
```
You are right. The problem refers to the SDK version in gradle project for android. 

You can fix this by modifying our project’s build.gradle file ([Android folder](https://github.com/RIP-Comm/sossoldi/tree/main/android)) and adding the ‘compileSdkVersion’ field.

Once you have updated the ‘compileSdkVersion’, you should also update any other dependencies that require a minimum SDK version of 33 or higher to ensure that our app functions correctly.

After doing this you can run: 
1. `git add --all` 
2. `git commit --amend` 
3. `git push --force` 
to update the failed commit.

theperu:
I changed compileSdkVersion but it seems that we still have problems, I am not able to really figure it out though. When I use it on my machine it seems fine and I don't have issues so it's also quite hard to understand what to change

FedericoBruzzone:
> I changed compileSdkVersion but it seems that we still have problems, I am not able to really figure it out though. When I use it on my machine it seems fine and I don't have issues so it's also quite hard to understand what to change

I understand, let's try this last test and if it doesn't work, we'll go back and put back what was written: `flutter.compileSdkVersion`

 Try to update in this way: ext.kotlin_version = '<latest-version>'

FedericoBruzzone:
LGTM :+1:

Issue #96, opened 2023-05-15
Overflowed text


mikev-cw:
On phisical iPhone 13 Pro, some texts on "settings" and "graphs" pages, is overflowing.
![IMG_3654](https://github.com/RIP-Comm/sossoldi/assets/113628339/7c6b3d84-e4f7-41c5-baa3-7895f01f1a86)
![IMG_3653](https://github.com/RIP-Comm/sossoldi/assets/113628339/02f601b9-4943-4167-8bf4-d80a4f49847e)

mikev-cw:
Graphs page overflow fixed on #110

0xbaggi:
I think we can close this issue because it has been solved

theperu:
This is also related to the widget in the graphs page, we will close it when the issue #49 will be closed.

PR #100, opened 2023-05-15
"active" field in bankAccount

mikev-cw:
- added "active" field in bankAccount Table
- Edited query to select all (active) bank Accounts
- Edited behavior of "Delete account" button, now just sets the account as inactive
- Fixed text overflow in settings page as per #96
- Imports cleanup, and general improvements

Delete your databases before testing.

### DIFF
--- a/lib/custom_widgets/accounts_sum.dart
+++ b/lib/custom_widgets/accounts_sum.dart
@@ -1,6 +1,6 @@
 import 'dart:ui';
 import 'package:flutter/material.dart';
-import 'package:sossoldi/constants/constants.dart';
+import '../constants/constants.dart';
 import '../model/bank_account.dart';
 import '../constants/functions.dart';
 import '../constants/style.dart';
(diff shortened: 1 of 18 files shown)


theperu:
Good work! Seems all fine, I just noticed one thing. If I try to add a transaction and I set a date in the past (yesterday or before) then I don't see it. No errors or warnings appear on the debug console. Maybe it is something that was already there and we didn't notice, I don't remember if I tried it before

lucaantonelli:
> Good work! Seems all fine, I just noticed one thing. If I try to add a transaction and I set a date in the past (yesterday or before) then I don't see it. No errors or warnings appear on the debug console. Maybe it is something that was already there and we didn't notice, I don't remember if I tried it before

About the transaction not showing up, maybe it's only because on the home page we show only the last 5 transactions. If you try to delete the today's transactions, you should see the older one.

mikev-cw:
Yeah, I think we need to speak A LOT about transactions... 😄 
I'm merging this for now.
Thank you!

PR #102, opened 2023-05-21
Improved collapsible header on Transactions page

### LINKED ISSUE #97: not visible text in calendar view
In "Transaction" page, when you want to select a time period clicking on the selector, the upper part of the page shows white on light grey text.
IMG_3656

GBergatto:
Fix #97 not visible text in Calendar view 

Date range is correctly displayed when the header is collapsed. Previously, it was hard-coded to "September 2022".

Start and end dates are now stored inside `transactions_page.dart`. This way they will be accessible by all tabs to be used for queries to the database.

Created `utils` folder for extension methods and utils functions.


FedericoBruzzone (approved):
Good job! It is okay for me.

LGTM :+1:

Issue #104, opened 2023-06-03
Make all items in dashboard use real data from DB


mikev-cw:
- [x] Main graph (Linked to #89)
- [x] Budgets graphs
- [x] accounts summaries widgets

PR #105, opened 2023-06-03
Dashboard account summaries fix

mikev-cw:
Accounts summaries in Dashboard now shows real data from database stored transactions (contributes to #104)
Improved demo data algo
Modified value field in bankAccount table to "starting_value"

### DIFF
--- a/lib/model/bank_account.dart
+++ b/lib/model/bank_account.dart
@@ -1,3 +1,4 @@
+import 'package:sossoldi/model/transaction.dart';
 import 'package:sqflite/sqflite.dart';
 
 import '../database/sossoldi_database.dart';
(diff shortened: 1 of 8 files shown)


lib/model/bank_account.dart
```diff
@@ -10,7 +11,7 @@ class BankAccountFields extends BaseEntityFields {
   static String name = 'name';
   static String symbol = 'symbol';
   static String color = 'color';
-  static String value = 'value';
+  static String starting_value = 'starting_value';
```
lucaantonelli, line 14: Is better to use camel case for variable names: "The variable name 'starting_value' isn't a lowerCamelCase identifier.
Try changing the name to follow the lowerCamelCase style." [https://dart-lang.github.io/linter/lints/non_constant_identifier_names.html](url) . This should be done in every file it is used.

mikev-cw: Damn... you're right. I will

lucaantonelli (approved):

theperu (approved):
Seems all good to me, we can merge if we won't make the update for the recurring transaction in the near future.

Issue #79, opened 2023-03-16
Refactor Planning page


theperu:
Since the restructure of the DB the current Planning page is not working properly anymore. I'm not 100% sure but I thing last time @0xbaggi worked on it. Not sure on how much work still needs to be done there so if there are more information that would be great for whoever is going to work on this issue.

SuperCaliMan:
I can check it

0xbaggi:
I'm working on it

mikev-cw:
Merged #107 for frontend. Backend functions needed yet, so i'm keeping this issue open for now

theperu:
@mikev-cw if you agree I would also close this issue since we introduced the budgets with #158 and the recurring payments are covered by #144

mikev-cw:
> @mikev-cw if you agree I would also close this issue since we introduced the budgets with #158 and the recurring payments are covered by #144 

Yeah, go on!

PR #107, opened 2023-06-30
Refactor Planning page

0xbaggi:
I started implementing the planning page with static data as requested in #79.
I will be ready for implementation of all feature when all data will be available from DB.

Screenshot_20230630_120352 (1)

https://github.com/RIP-Comm/sossoldi/assets/39710037/4b95c3ef-4e39-447c-89a4-65d7a0f3fc49


GBergatto:
Good job! I'm not sure about the effect when pressing on the pie chart. I would make it expand less, to keep it consistent with the pie chart inside the transactions page. Click [here](https://github.com/RIP-Comm/sossoldi/blob/main/lib/pages/transactions_page/widgets/categories_pie_chart.dart) for the code. Also, I don't think we need to show the exact percentage when pressing on a sector.

GBergatto:
Also, I don't think this PR should change `pubspec.lock`. BTW please use more expressive commit messages :)

0xbaggi:
> Good job! I'm not sure about the effect when pressing on the pie chart. I would make it expand less, to keep it consistent with the pie chart inside the transactions page. Click [here](https://github.com/RIP-Comm/sossoldi/blob/main/lib/pages/transactions_page/widgets/categories_pie_chart.dart) for the code. Also, I don't think we need to show the exact percentage when pressing on a sector.

Thank you! I will disable the effect because if I remove the percentage it becomes useless,

GBergatto:
Instead of having two commits followed by one to revert their effect, you should delete the first two all together. I believe this can be done using `git reset`. However, before you do, maybe ask someone with more experience to make sure it won't cause any damage.

mikev-cw:
Hello, thank you for this work!!
I agree with Gbergatto's excellent advices. I also agreed in keeping any kind of consistency with the pie chart on the transaction page.
For all edits mentioned you want to update this PR, or prefer to merge and open a new one?

I would like to further emphasize what has already been mentioned: Please remember to get in PRs only files inside `assets`, `lib` and `test` folders, and to use better commit descriptions. :)

0xbaggi:
> Hello, thank you for this work!! I agree with Gbergatto's excellent advices. I also agreed in keeping any kind of consistency with the pie chart on the transaction page. For all edits mentioned you want to update this PR, or prefer to merge and open a new one?
> 
> I would like to further emphasize what has already been mentioned: Please remember to get in PRs only files inside `assets`, `lib` and `test` folders + files flutter related like pubspec.yaml, and to use better commit descriptions. :)

I updated this PR with all the changes completed. I think it’s ready for the merge.

PR #109, opened 2023-07-10
Starting balance field in account creation page

### LINKED ISSUE #106: Add starting_value when adding accounts
In the "add account" widget, we need also the starting account value (which now is writed as 0)

Paxol:
I added a numeric field for the `startingValue` field in the account creation page, this will close #106.
It's my first time working with flutter, so tell me anything wrong 😅


mikev-cw:
Thank you for your contribution!!! I agreed with @theperu, we'll check for that details later.
I think this is ready for merge, but this is a draft PR and cannot do that. 
@Paxol Do you think is ready to turn it into a regular PR?

Issue #39, opened 2022-12-19
Onboarding


mikev-cw:
We need an onboarding funnel as seen on [Figma](https://www.figma.com/file/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?node-id=1163%3A4030).
Please, if you want to work on this, let us know leaving a comment below!

theperu:
@0xbaggi if I am not mistaken you were working on this, are you still going forward with it? You can post here or on Discord if you have any problem

mene412:
I would like to work on it

mikev-cw:
I'm closing this, since we managed to get the onboarding done on #162

PR #114, opened 2023-08-15
Onboarding - first part

mene412:
Issue #39
TODO:
- Make the onboarding page visible only on the first visit.
- Connect the "ADD BUDGET" feature and the account_setup.dart page with the database.
- Validate the presence of data when the "START TRACKING YOUR EXPENSES" button is pressed.


FedericoBruzzone (approved):

theperu:
Hi @mene412! Do you think that you could also work on the parts that you pointed out are missing in order to complete this feature? In that case we might want to keep this open

mene412:
Up to today, I am unable to proceed because a part of the provider is missing, specifically `budgets_provider.dart`, in order to add the budgets. I tried to modify the file, but I haven't reached a solution yet.

theperu:
@mikev-cw and @lucaantonelli what do you think? Do we want to merge this now or you have suggestions/time to help mene on the next steps?

mikev-cw:
I see that this was originally a `draft` PR, converted to `ready for review`, then approved, for reasons I personally don't see. Now has even unresolved conflicts.
So I don't think this is ready to merge: Onboarding is a critical and primary feature that needs to be as ready as possible to merge (also to not hinder development, imagine having to do an onboarding that doesn't work every time the app is launched).
So @mene412 if there's something missing, as budget provider you mentioned, tell us if you can work on it, or you need someone else to help. Whatever the case, i think we should put this into an issue.

Issue #70, opened 2023-02-02
Currency selector on Settings


theperu:
As discussed in the last call we should add the option to change the currency used in the app.
I think that putting this option is enough and for now I don't think it is necessary to also add it to the Onboarding. We can set the euro as the default and the add the option for the most important for Mr.RIP followers (Dollar, Pound and Swiss Franc).

mikev-cw:
To anyone who will take charge on this: Currencies list should be taken from DB, I'm working on that table.

AntonioMiclaus:
If nobody is working on it, I can start the task.
Currently we have four main categories in the settings page:  Accounts, Categories, Budget, Import/Export.
Should I add one more with general app options? Like this one about the currency used, the light/dark mode, etc.

theperu:
Hi @AntonioMiclaus! How are things going with this issue? This is not to put pressure, we are going to have a meeting in the near future and I would like to understand were we are.

theperu:
Ok, just to clarify at the moment we have a currency selector (settings -> general settings -> currency) but it does nothing when you select a different currency. What we need to implement is a place in which we save the selected currency and always display that instead of the standard "€" that we are doing right now

PR #115, opened 2023-08-26
Start implementation of general settings

0xbaggi:
As suggested in #70, I added a new section (general setting) in the page settings, which will have general settings like light/dark mode and currency selector. I haven't started to implement the page because it hasn't been defined on Figma yet.
image


FedericoBruzzone (approved):

FedericoBruzzone (approved):

theperu:
Hi @0xbaggi ! Since I haven't heard back from the guy who was working on #70 if you want you can start implementing that part in the general settings. Personally I wouldn't wait for the figma since this is a small part of the app and there shouldn't be many problems with it. 
For the dark mode I remember that @napitek already tried something but we can implement it in the future.

0xbaggi:
I added some static buttons in the general settings page, their functionalities have yet to be implemented.

- Appearance to enable/disable dark mode
- Currency to select your preferred currency
- Language to select the app’s default language

[Screen_recording_20230913_232758.webm](https://github.com/RIP-Comm/sossoldi/assets/39710037/a8f9c9b5-fee3-47d5-8fdf-c70e0c4fb446)

theperu (changes requested):
Hi I would remove the language picker (or we can keep the code and just avoid displaying it) for now since we won't support other languages before version 1.0.
I see that the check failed with this message:

> The current Dart SDK version is 2.18.0.
> Because sossoldi depends on country_flags >=2.1.0 which requires SDK version >=3.0.0 <4.0.0, version solving failed.

We should probably look into that, we might think about using the emojis for the flags (Ex. 🇮🇹, 🇬🇧, 🇩🇪, 🇫🇷) instead of something external.

0xbaggi:
I removed the library, replaced the flags with the relative emojis and hid the language setting for future versions.
The hidden part is made with emoji without the library(not currently visible because commented):

image

theperu (approved):

theperu (approved):
Looks good and the check passed, @mikev-cw if you are ok with that I think we can merge this

mikev-cw:
@theperu if you already checked, it's enough. Let's merge!

Issue #108, opened 2023-07-04
Refactor widget code


mikev-cw:
Some widgets are now builded from scratch every time.
We need to standardize this part of the codebase

GBergatto:
I'll start doing some work related to this issue while working on the ex Account modal for issue #90

GBergatto:
Welcome! Yes, there are some widgets that are shared across multiple pages but are created from scratch on each of them. The goal is to define widgets that we can reuse across the page to make the code more organized and avoid having to rewrite the same things over and over. An example, is `lib/custom_widgets/transactions_list.dart` which I've worked on in my last PR #118.

giuseppe-coco:
Hello, sorry for my late response but I'm currently working and haven't had much time to devote to the project.

I noticed some strange things that did not allow me to understand in detail the functions of the app and the related widgets that control them, so I could not make changes.

Specifically , the app cannot find the "Main Account" .

<img width="200" alt="graphs" src="https://github.com/RIP-Comm/sossoldi/assets/76009241/ed31a446-c699-4e23-869e-751faf43c2ed">
<img width="197" alt="home" src="https://github.com/RIP-Comm/sossoldi/assets/76009241/5a7e40e4-b9fa-4415-9f89-c8d8c274920b">
<img width="209" alt="transactions" src="https://github.com/RIP-Comm/sossoldi/assets/76009241/0def5115-89c9-4c9e-adc5-dad0d653394e">

GBergatto:
@giuseppe-coco I think that that happens because you don't have any transactions, yet. Try to go to "Settings" and, in the "dev only" zone, do "clear and fill demo data". This will create a bunch of random transactions in the local database

giuseppe-coco:
Yes that was the problem, thank you!

PR #118, opened 2023-10-05
Page to add/edit transactions and account page

### LINKED ISSUE #90: Transform accounts modal to a page
Currently the lower part of every account modal shows some weird things instead of the right transactions. I think we can re-use a lot of the work that as been done for the transaction page by @GBergatto (#84 )
[Figma link for reference](https://www.figma.com/file/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?type=design&node-id=3014-7173&t=ly7oFThllyoSddw1-4)

EDIT: After discussing this issue on discord we decided to remove the modal and use a page instead

### LINKED ISSUE #98: not visible field while filling
In the "add new transaction" modal, "notes" field is not visible while filling

https://github.com/RIP-Comm/sossoldi/assets/113628339/efbbac22-adac-450b-a98b-9c50b389cdff

GBergatto:
I've transformed the modal used to create/add transactions into a page (fix #90), called `AddPage`. I've also put the text for the label right after the price to fix #98. Finally, I've made some improvements to the list of transactions in the "List" tab of the "Transactions" page: the daily totals are now correct and the icons are colored based on the category of the transaction.

In addition, I've transformed the modal that used to open when clicking on the cards in the "Your accounts" section of the homepage into a page. However, I haven't added the list of transactions to it, yet, and the total amount displayed on the page doesn't match the one on the card.

WARNING:
Clicking on the transactions in the "Last transactions" section of the page currently opens a blank modal.

I haven't edited this section because my plan is to replace it with the `lib/pages/transactions_page/widgets/list_tab.dart` widget, after having made it more versatile so that it can be reused into different parts of the app. This include adding a way to filter transactions by account (to use it in the account page) and to limit the number of transactions displayed (to use it in "Last transactions" on the homepage).
This is in line with #108 about reusing widgets.

If that's fine with you, I would like to merge this PR to keep momentum and avoid big merge conflicts down the line. Next, I'll work on fixing the issues mentioned.


lucaantonelli:
@GBergatto I've checked the PR, the new add transaction page looks good, but update fail since use the old modal page, we should update the onTap at row 189 `showModalBottomSheet(...` in _custom_widgets\transactions_list.dart_ , just changing row 188 to that should do the work (removing the modal code block):
`ref.read(transactionsProvider.notifier).transactionUpdateState().whenComplete(() => 
  Navigator.of(context).pushNamed("/add-page"));`

lib/pages/add_page/widgets/recurrence_list_tile.dart
```diff
@@ -0,0 +1,140 @@
...
+            ),
+          ),
+          title: Text(
+            "Recurring payment",
+            style: Theme.of(context)
+                .textTheme
+                .titleLarge!
+                .copyWith(color: Theme.of(context).colorScheme.primary),
+          ),
+          trailing: (Platform.isIOS)
```
lucaantonelli, line 46: In dart exists some great adaptive widgets, switch is one of that, we should move to Switch.adaptive(), it will handle automatically the platform, using less code.

GBergatto: Thanks, I'll change it to that

GBergatto:
@lucaantonelli I've just applied the changes you suggested. Now, clicking on a transaction list tile on the homepage will open the page to edit it instead of the empty modal.

GBergatto:
There's a noticeable delay when trying to switch to the homepage from another page. My guess is that having to load data for the graph, account sums, and transactions slows things down. Maybe this should be investigated in more details in the future

lucaantonelli:
I noticed some strange behaviour when changing amount field (in update, but something broken also when creating new transactions), 
Apart from that LGTM.
281688023-3f0eb3ea-abcb-4445-aa44-3433a9748a4e.mp4

GBergatto:
@lucaantonelli It has something to do with the decimal separator switching from comma to dot. I'll look into it

mikev-cw:
> @lucaantonelli It has something to do with the decimal separator switching from comma to dot. I'll look into it

@rcasula worked (hardly 🤣) on that months ago. See if he can possibly help, if you need.

> There's a noticeable delay when trying to switch to the homepage from another page. My guess is that having to load data for the graph, account sums, and transactions slows things down. Maybe this should be investigated in more details in the future

Not detected honestly. Please explain what "noticeable" means to you

rcasula:
I've tried it and LGTM! Great work

GBergatto:
@mikev-cw I checked and I don't think it has something to do with the decimal separator (fortunately) because it happens even when the separator doesn't change. I guess it's caused by the way we update the provider.

About the delay, it's last just a split second. If you want I can send you a screen recording of it. Consider that I've edited `lib/providers/transactions_provider.dart` to change the hard-coded limit of transactions from 5 to 50. So, maybe that has an impact on it.

mikev-cw:
> About the delay, it's last just a split second. If you want I can send you a screen recording of it. Consider that I've edited `lib/providers/transactions_provider.dart` to change the hard-coded limit of transactions from 5 to 50. So, maybe that has an impact on it.

I've tried with 100 transactions, and still not noticed any remarkable delay.
If you want, after fixing that strange behavior with the amount field (and resolve new conflicts, sorry for that!), send us a screen recording, and we'll try to go deeper on that!

lucaantonelli:
LGTM, except known bugs it works fine.

mikev-cw:
Cool!
Let's merge

PR #122, opened 2023-11-25
add beta light/dark mode 1.1

giovanni-ercolano:
I made some changes from the previous pull request, in particular I corrected what I had been told previously, I hope everything is correct now. I specify that obviously not all pages have an exact implementation of dark mode but it is due to the fact that they are still under development of course. 

Ps. I tried to insert only the lib folder as you requested, but it was done only with the last commit, from next time I will start from the beginning.


mikev-cw:
Hello, and thank you for your contribution!
I've pushed a commit to revert most system files. Please be aware of this in the future. 
If you want, we would really appreciate if you could monitor all dark mode stuff while developing goes forward, and fix possible problems.

Thank you, bye!!

giovanni-ercolano:
Hi, I'm very happy that my work went well since this was my first open source contribution. That said, I will definitely continue working on the project in the future.
Thanks again everyone, bye.

PR #125, opened 2023-12-09
New docs for the creation of a GitHub Page

theperu:
As discussed in the Discord, I experimented a little bit with GitHub Pages and I think it is the best way for us to have our documentation somewhere, we can also use it has a landing page when people will ask about the project (once is ready I'll ask Mr.RIP if he can create a short link for it). You can see an example of how it will look here: https://theperu.github.io/sossoldi/

**How it works**
It is pretty straightforward, we can just use .md files and then GitHub will do the conversion to HTML for our website. The main folder would be the _docs_ and it will use the _README.md_ inside it as the main page
In the future we could also make it prettier using some of the themes that are available or by doing something custom


theperu:
@mikev-cw or @lucaantonelli if you're ok with it can you merge this so that I can create the GitHub page?

mikev-cw:
Thank you!
Not sure why tests are failing using Flutter SDK 3.1.0... we require >3.1.5... 🤔 
Other than that, it's ok to me

mikev-cw (approved):

theperu:
I had the same doubt, I don't think that the changes that I made caused the failure because they're just images and .md files

lucaantonelli:
@mikev-cw from the error it suggests "Try using the Flutter SDK version: 3.16.3", we are using 3.13.9 in the workflow, so i think that could be the problem

theperu:
As discussed in the Discord I'll merge this PR

Issue #112, opened 2023-08-02
Localised decimal separator


rcasula:
Use the system default decimal separator instead of the hardcoded `.`.

GBergatto:
Currently, we are using commas and periods inconsistently across the app. The best way to solve this is, as suggested by @rcasula, to read the local of the user. However, in this early stage we could start with the English locale as default, using dot as decimal separator and comma as thousands separator.

theperu:
I agree with @GBergatto since the app is in English I think we should stick with `.` for the decimals and `,` for the thousands

Votyer:
i'll take care of it

Votyer:
Hello everyone, short update from my task.
I changed the method "numToCurrency" in functions class so now it accept the locale as parameter. 
Addictionally I added the flutter lib for string internationalizing(link below)  to support different languages.

I hope to close this task before the end next week

Internationalizing Flutter apps
https://docs.flutter.dev/ui/accessibility-and-internationalization/internationalization

theperu:
Great! We'll wait for the PR then!

Votyer:
PR Opened sorry for delay!

PR #128, opened 2023-12-19
Feature: Added flutter internationalizing

Votyer:
#112


l10n.yaml
```diff
@@ -0,0 +1,3 @@
+arb-dir: lib/l10n
+template-arb-file: app_en.arb
+output-localization-file: app_localizations.dart
```
theperu, line 3: When I run the main I get an error because it looks like that the word dashboard doesn't have a translation in the Italian file, you can check the missing words by adding here `untranslated-messages-file: missingWords.txt`

theperu (changes requested):
Hi! I tried to run your changes and I noticed a few things. It seems to me that only a few things are translated while others are always kept in english which makes the app look a bit weird, some of these things are: dates, graphs page, "planned" and "planning" inside the Planning page, "Transactions" in the Transactions page. 
Apart from this I tested the app on Android and it looks like the "Add transaction" page as some issue and throws an error (see the video below)

https://github.com/RIP-Comm/sossoldi/assets/24766687/92485691-b572-4073-aab2-773ffd6da058

In the end I think that going forward with a complete translation of the app in Italian might be good but it will cost quite some time since we have smooth a lot of things out. I'll wait also for the opinion of @mikev-cw and @lucaantonelli

mikev-cw:
Same error of @theperu on the "Add" page.

> In the end I think that going forward with a complete translation of the app in Italian might be good but it will cost quite some time since we have smooth a lot of things out. I'll wait also for the opinion of @mikev-cw and @lucaantonelli

Anyway I'm agreeding with this. This work is good enough and have all the "base things" we need, but it threatens to take away time from the milestones. Maybe we can speak on Discord about keeping this and go deeper, or park somewhere.

theperu:
@mikev-cw and @lucaantonelli the support for other languages is quite far in the future in my opinion so I think we should just close this and comeback to it once the app is more mature and stable. 
@Votyer the work that you have done is interesting and it might be useful in the future but unfortunately I think we need to focus on our milestone and releasing the beta now, I really appreciate your work though and I hope you'll keep helping the project going forward 🙂

mikev-cw:
> @mikev-cw and @lucaantonelli the support for other languages is quite far in the future in my opinion so I think we should just close this and comeback to it once the app is more mature and stable. @Votyer the work that you have done is interesting and it might be useful in the future but unfortunately I think we need to focus on our milestone and releasing the beta now, I really appreciate your work though and I hope you'll keep helping the project going forward 🙂

100% agreed.
This will be our starting point when this task becomes higher on priority. Thank you!

PR #129, opened 2023-12-21
Transactions date filtering + settings improvement

### LINKED ISSUE #88: transfers addition issue
I am able to submit a transfer without the second account selected, when I comeback to the home then I see an error in the last transactions section.

https://user-images.githubusercontent.com/24766687/235300185-12ce5734-b874-4760-983d-081e6256ff2f.mov

(From a comment of @theperu on #85)

lucaantonelli:
- Add notification settings
- Update layouts and color to match figma
- Add filter for month on transactions tab with correct design on list
- Merge of transaction_list_tile.dart in transaction_list.dart to have only one widget
- Working create/update/delete of transactions
- Partial budget page improvement
- close #88


theperu (changes requested):
Great work! I just one question and two small suggestions/requests. I saw that you added the notification toggle but we still need to make it work right?
For the suggestions:
-  I see that the bug for the transfer is fixed but a user is still able to make a transaction from an account to the same one (ex. Revolut -> Revolut) which is a bit strange to me, it would be better if we could avoid this kind of behaviour. Of course it depends on how much time this change would take, if it's too much let's leave it as is.
- The word "View:" in the Transactions page looks odd to me and seems unnecessary, I see it also in the Figma from @federicopozzato but if he agrees I would remove it since it should be clear to the user what they are seeing in that page (I know that this was already there but I just noticed and we could take advantage of this PR to remove it)
Screenshot_2023-12-24-12-58-34-741_com example sossoldi

federicopozzato:
@theperu agreed👌

lucaantonelli:
@theperu i updated the code following your suggestions, and yes, actually the notifications settings is just graphically done.

mikev-cw:
Super!!
I see some errors on Categories and Accounts tabs (the old bad boy `"type 'int' is not a subtype of type 'double'"`), but most probably are just TODOs.

Other than that looks very good to me. If all of you agreed, we can surely merge.

lucaantonelli:
@mikev-cw Thanks! About the error, it might be about data, clearing in settings should solve it.

lucaantonelli:
I just did a lot of work on the PR!
Now the dashboard is working, updated some queries with join to use less code inside widgets, categories and accounts tab in transactions is working, even the selection between income and expense (probably not the best way, but for now it's working great).

PR #133, opened 2024-01-05
Issue 89

PietroChitto:
* Reading real data in account page graph
* Reading real data in graph page graph

### DIFF
--- a/lib/pages/account_page/account_page.dart
+++ b/lib/pages/account_page/account_page.dart
@@ -18,6 +18,7 @@ class _AccountPage extends ConsumerState<AccountPage> with Functions {
   @override
   Widget build(BuildContext context) {
     final account = ref.read(selectedAccountProvider);
+    final accountTransactions = ref.watch(selectedAccountCurrentMonthDailyBalanceProvider);
 
     return Scaffold(
       appBar: AppBar(
@@ -43,29 +44,11 @@ class _AccountPage extends ConsumerState<AccountPage> with Functions {
                     ),
                   ),
                   const Padding(padding: EdgeInsets.all(8.0)),
-                  const LineChartWidget(
-                    line1Data: [
-                      FlSpot(0, 3),
-                      FlSpot(1, 1.3),
-                      FlSpot(2, -2),
-                      FlSpot(3, -4.5),
-                      FlSpot(4, -5),
-                      FlSpot(5, -2.2),
-                      FlSpot(6, -3.1),
-                      FlSpot(7, -0.2),
-                      FlSpot(8, -4),
-                      FlSpot(9, -3),
-                      FlSpot(10, -2),
-                      FlSpot(11, -4),
-                      FlSpot(12, 3),
-                      FlSpot(13, 1.3),
-                      FlSpot(14, -2),
-                      FlSpot(15, -4.5),
-                      FlSpot(16, 2.5),
-                    ],
-                    colorLine1Data: Color(0xffffffff),
-                    line2Data: <FlSpot>[],
-                    colorLine2Data: Color(0xffffffff),
+                  LineChartWidget(
+                    line1Data: accountTransactions,
+                    colorLine1Data: const Color(0xffffffff),
+                    line2Data: const <FlSpot>[],
+                    colorLine2Data: const Color(0xffffffff),
                     colorBackground: blue5,
                     maxDays: 30.0,
                   ),
--- a/lib/pages/account_page/account_page.dart
+++ b/lib/pages/account_page/account_page.dart
@@ -44,13 +44,16 @@ class _AccountPage extends ConsumerState<AccountPage> with Functions {
                     ),
                   ),
                   const Padding(padding: EdgeInsets.all(8.0)),
-                  LineChartWidget(
-                    line1Data: accountTransactions,
-                    colorLine1Data: const Color(0xffffffff),
-                    line2Data: const <FlSpot>[],
-                    colorLine2Data: const Color(0xffffffff),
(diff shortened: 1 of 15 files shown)


lib/pages/account_page/account_page.dart
```diff
@@ -43,29 +44,11 @@ class _AccountPage extends ConsumerState<AccountPage> with Functions {
...
-                      FlSpot(13, 1.3),
-                      FlSpot(14, -2),
-                      FlSpot(15, -4.5),
-                      FlSpot(16, 2.5),
-                    ],
-                    colorLine1Data: Color(0xffffffff),
-                    line2Data: <FlSpot>[],
-                    colorLine2Data: Color(0xffffffff),
+                  LineChartWidget(
+                    line1Data: accountTransactions,
```
theperu, line 48: It looks to me like the graph is not showing the data that I was expecting, when I load the demo data and check the Revolut account for example I see a negative value. What I think the graph should show is how the account balance changed in the past month day-by-day (let me know if I wasn't clear and I'll write an example)

lib/pages/account_page/account_page.dart
```diff
@@ -43,29 +44,11 @@ class _AccountPage extends ConsumerState<AccountPage> with Functions {
...
-                      FlSpot(12, 3),
-                      FlSpot(13, 1.3),
-                      FlSpot(14, -2),
-                      FlSpot(15, -4.5),
-                      FlSpot(16, 2.5),
-                    ],
-                    colorLine1Data: Color(0xffffffff),
-                    line2Data: <FlSpot>[],
-                    colorLine2Data: Color(0xffffffff),
+                  LineChartWidget(
```
theperu, line 47: We should probably show some days at the bottom of the chart as a reference. See the [Figma](https://figma.com/file/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?type=design&node-id=0-1&mode=design&t=t2tpmLbyvuLyOtRZ-0)

theperu (changes requested):
Wow that was fast! I checked your changes and it looks to me like that there are only a few things to modify. I commented directly for the specific things and then there is one that affects both graphs and it's how they are displayed, at least on my android device they start directly from the edge of the screen which makes them hard to interact and read. 
See the screenshoots:
Screenshot_2024-01-05-19-01-26-736_com example sossoldi 1

Screenshot_2024-01-05-19-01-18-186_com example sossoldi 1

About the Graphs page: since we are in January it's not showing anything and my guess is because there is only one point in the graph, what we could do is make the graph go from December to December or find another graphical way to display what's the liquidity for this month

lib/pages/account_page/account_page.dart
```diff
@@ -43,29 +44,11 @@ class _AccountPage extends ConsumerState<AccountPage> with Functions {
...
+                    line1Data: accountTransactions,
+                    colorLine1Data: const Color(0xffffffff),
+                    line2Data: const <FlSpot>[],
+                    colorLine2Data: const Color(0xffffffff),
                     colorBackground: blue5,
                     maxDays: 30.0,
```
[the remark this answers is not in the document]
PietroChitto: I agree

lib/pages/account_page/account_page.dart
PietroChitto: This graph is like that in the dashboard filtered by account. If we want to take the balance day by day considering all the history, we must also take into account the transfer transactions from one account to another and we need to understand how to calculate the balance efficiently.

Suppose this situation:
-Inital amount of account A: 1000
-Initial amount of account B: 1000
- expense of 100 on 31th Dec on A
- transfert 50  from A to B on 1st Jan
- Income of 100  on 2nd Jan on A

The points for the two account for Jan should be:
A: (1, 850), (2, 950)
B: (1, 1050), (2, 1050)

Can you confirm this?

Another point is how to calculate the balance, now the only possible way is to calculate it strating from day 0 and considering all the transacions. Until there are few transactions this has a limited calculation impact (fine for a first implementation) but when there are a significant number of transactions this could have a considerable impact (to understand how it grows as transactions increase) and alternative will need to be considered.

lib/pages/account_page/account_page.dart
PietroChitto: I agree, it is relative to the widget

lib/pages/account_page/account_page.dart
theperu: Yes, that was exactly what I had in my mind. About the balance, in the past we talked a little bit on how to do it so that we don't have to do many operations and we considered both having some kind of record inside the DB (ex. every month we save the balance) or use the cache but we didn't finalize our decision

PietroChitto: Ok, fine, now I implement it by recalculating it every time. If there will be records with balance in db in the future let's remember that every time an update transaction is performed, the monthly balances following the transaction must be recalculated

theperu (commented):
Everything seems fine to me except one thing that I noticed in the account page, it seems to me that the chart there has the minimum of the List that we send as the bottom of the chart instead of zero. You can see the screenshot below in which it looks like my account should be empty but I have 38k.

Screenshot_2024-01-14-17-34-26-543_com example sossoldi 1

PietroChitto:
The same behviour is present in all line charts in the application (Dashboard, account and graph pages). I propose to collect all these requirements and create a separate issue.
Now we have:

- [ ] Manage tooltip that go out of the screen
- [ ] Set a minimum y value
- [ ] Find a better way to display a single point in l;ine chart (now nothing is shown)

PR #134, opened 2024-01-18
fix linechart

PietroChitto:
* improved linechart tooltip
* show message when there are 0 or 1 point
* tooltips don't exit the screen anymore
* added bar indicator
* add possibility to set min y


lib/custom_widgets/line_chart.dart
```diff
@@ -10,21 +18,38 @@ class LineChartWidget extends StatefulWidget {
   final Color colorLine2Data;
 
   //Contains the number of days of the month
-  final double maxDays;
+  //final double maxDays;
 
+  // Used to decide the bottom labal
```
theperu, line 23: This might be a typo 🙂

lib/custom_widgets/line_chart.dart
```diff
@@ -10,21 +18,38 @@ class LineChartWidget extends StatefulWidget {
   final Color colorLine2Data;
 
   //Contains the number of days of the month
-  final double maxDays;
+  //final double maxDays;
```
theperu, line 21: I think we can just remove this

theperu (approved):
Apart from two small things it looks good to me and we can merge it 😁

theperu (approved):

theperu:
I think we can merge this since the failed check is not related to the linechart, @mikev-cw or @lucaantonelli do you agree?

lucaantonelli (approved):

PR #141, opened 2024-01-31
Feature/search page

K-w-e:
Here my first version of the search page, i kept it simple.
You can select/deselect to search transactions by type or in a specified account. (every filter is already enabled on initial state)

Let me know if you have any suggestions!

### DIFF
--- a/lib/pages/search_page/search_page.dart
+++ b/lib/pages/search_page/search_page.dart
@@ -81,8 +81,10 @@ class _SearchPage extends ConsumerState<SearchPage> {
                         _updateFutureTransactions();
                       },
                     )),
+                Container(
+                    alignment: Alignment.centerLeft,
+                    child:const Text("SEARCH FOR:")),
                 Row(children: [
-                  const Text("Search for: "),
                   Expanded(
                     child: SingleChildScrollView(
                         scrollDirection: Axis.horizontal,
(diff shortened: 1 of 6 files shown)


lib/providers/accounts_provider.dart
```diff
@@ -11,11 +11,18 @@ final accountColorProvider = StateProvider<int>((ref) => 0);
...
 final countNetWorthSwitchProvider = StateProvider<bool>((ref) => true);
 final selectedAccountCurrentMonthDailyBalanceProvider = StateProvider<List<FlSpot>>((ref) => const []);
+final filterAccountProvider = StateProvider<Map<int, bool>>((ref) => {});
+
 
 class AsyncAccountsNotifier extends AsyncNotifier<List<BankAccount>> {
   @override
   Future<List<BankAccount>> build() async {
     ref.watch(mainAccountProvider.notifier).state = await _getMainAccount();
+    _getAccounts().then((value) => {
```
lucaantonelli, line 21: I think the _getAccounts() call is redundant, you can just put a return inside the first one after the for loop, it should work.

```
return _getAccounts().then((value) => {
      for(BankAccount account in value) {
        ref.watch(filterAccountProvider.notifier).state[account.id!] = true
      }
      return value;
});
```

lucaantonelli (changes requested):

theperu (commented):
Apart from what has been already pointed out it looks good to me. I would just leave it open a bit and see if Federico (UX/UI designer) has any suggestion, great work! 😁

federicopozzato:
Great work! I just revisited a bit under the UI point of view. You can find it in the figma (one of the last frames, 6.6.0), but I leave you a screenshot here: 
image

K-w-e:
PR updated
I'm adding one thing, the initial state of the filters is no longer the same. Now, initially, only Income/Outcome and the main bank account are selected. In my opinion, with the new interface, if everything was selected from the start, it wasn't clear.

Let me know!

federicopozzato:
That default state should be that none of the filters is applied, so all should be empty, then you can filter by tapping in them and when they are active they become filled with blue.

Was this your doubt?

K-w-e:
Oh ok, fixed it

theperu (approved):
LGTM, I think we can merge this 👍🏻

lib/pages/search_page/search_page.dart
```diff
@@ -0,0 +1,257 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:sossoldi/providers/accounts_provider.dart';
+import 'package:sossoldi/providers/transactions_provider.dart';
+import '../../custom_widgets/transactions_list.dart';
+import '../../model/transaction.dart';
```
mikev-cw, line 6: Can you fit the first 2 imports paths to our code guidelines?
Just do the same as the last 2 imports, with the absolute path

mikev-cw (changes requested):
LGTM too, just fix this small thing, then you're done.

Also I'm not sure about letting all disabled by default is the clearest thing, but we can speak about that on next meetings, for now it's great!

lib/pages/search_page/search_page.dart
K-w-e: Sure

Issue #140, opened 2024-01-29
Accounts list is not updating 


jackrua:
**To reproduce the issue** go to: 

`Settings` > `Your accounts` > `Add account` > (insert any name and value you want) > `Create account` > Go back and you won't see any new account. 

**What I have gathered so far**

This is a really interesting bug that brought me to discover some of the internal logic of the app, but I can't really get to the bottom of it. 

It seems to me that the `sossoldi.db` is actually updated when you trigger this action, however the `accountsList` variable (in the `_AccountListState` object) is not updated, so it doesn't show up in the widget. 

When you add an account, the stack goes like this: 

1. `accountsList` watches `accountsProvider` which is the `AsyncAccountsNotifier()` notifier
2. which in turn uses the `BankAccountMethods` object to interface with the db. 
Here everything seems to work (hence why the db is updated when you add the account)

So I would say that the issue lies in `1.` . However when you update an account everything works as it should, so in this class (`AsyncAccountsNotifier()`) `updateAccount()` works, while `addAccount()` doesn't. 

If anybody, more experienced than me, could point me in the right direction I would love to close this PR.

theperu:
Hi! If you want I can assign this issue to you while you try to figure out the root cause of the bug. Since the issue might go unnoticed for a bit if you want you can try asking the same question also in the Discord channel since it is a bit more active

jackrua:
Finally found it! It took me more than I would like to admit. But contrary to my first hunch, the issue was in the SQL query in the `selectAll()` method in `BankAccountMethods`. 
This method launches an SQL query that looks something like this: 

```
SELECT b.*, (b.startingValue +
      SUM(CASE WHEN t.type = 'IN' OR t.type = 'TRSF' AND t.idBankAccountTransfer = b.id THEN t.amount
               ELSE 0 END) -
      SUM(CASE WHEN t.type = 'OUT' OR t.type = 'TRSF' AND t.idBankAccount = b.id THEN t.amount
               ELSE 0 END)
    ) as total
      FROM bankAccount as b
      LEFT JOIN "transaction" as t ON t.idBankAccount = b.id OR t.idBankAccountTransfer = b.id
      WHERE active = 1 AND recurring = 0
      GROUP BY b.id
      ORDER BY createdAt AS
   ```
   
[...]

PR #146, opened 2024-02-04
Fixes in accounts list 

jackrua:
Two fixes:

**First fix** (the important one): 

* Addresses issue #140. 
* The problem is already illustrated at the issue page, I tried to make as little changes as possible to the original code. Now when you query `sossoldi.db`, you also get the accounts that have no transactions. 

**Second fix**: 

* Added a little padding since the "Delete Account" button is partially covered on Pixel 5 (and completely covered on my personal phone) by the "Update Account" button. See bottom of picture below:


jackrua:
My bad! I just saw that the fix has been already implemented ;)

Issue #147, opened 2024-02-04
Frequent categories are not clickable 


jackrua:
**To reproduce the issue**: 

Add a new transaction > choose category > the categories under "MORE FREQUENT" are not clickable 
(the same goes when you choose an account) 

A screen below: 

![p3](https://github.com/RIP-Comm/sossoldi/assets/155536850/57fe347d-3021-4257-8fc9-38898f4d71a4)

PR #148, opened 2024-02-04
fix on tap more frequent catogories/accounts

JokerZenDev:
Fix issue #147 adding a GestureDetector to handle on tap event.


jackrua:
Great! The code seems fine to me and it works  on my end.

mikev-cw:
Agreed! 
Thank you and welcome, let's merge

Issue #149, opened 2024-02-05
Adding transaction list to accounts 


jackrua:
I'm writing this issue to track this feature I would like to add. 

In the accounts page (under the line chart) we would like to show the transactions associated to each account.

 
![p11](https://github.com/RIP-Comm/sossoldi/assets/155536850/82c5ee2c-58c2-44c2-8f7c-33dc028c4166)

(this also addresses TODO comment at line `62` of `pages/account_page/account_page.dart`)

PR #152, opened 2024-02-11
Adding transactions to account page

jackrua:
(Addressing issue #149)

**Disclaimer**: 

You may notice more changes in files than there are actually, since I just noticed that my Editor automatically formats code so in some cases lines are just reformatted and nothing is really changed. Hence, I will try to be specific on the main changes I made to the code. However if you feel like this is an issue I can close this PR and open a new one. 

**Overview**:

To address this issue, we need to work with both the frontend of the app (i.e. find the correct widget to represent the transactions)  and its backend (i.e. retrieve from the database the transactions from a specific bank account and define the providers that the widget can watch). So, moving from bottom to top of the stack, what I did was: 

* In commit [fe9eec3](https://github.com/RIP-Comm/sossoldi/commit/fe9eec376cb3866c358e92951487fea9e9de154f) I added a method to `bankAccountMethods()` so that I can interrogate the db to get the last n transactions from a particular account 

* In commit [a3c7714](https://github.com/RIP-Comm/sossoldi/commit/a3c7714d5c58f43837bb75640ce1a12035ce1dea) I defined the provider (`selectedAccountLastTransactions`) that will be exposed to the account page widget to update the transaction list. 
**Note**: Here I hardcoded that the widget will retrieve just the last 50 transactions (due to efficiency reasons). This is ugly, I admit, but for the time being it should be ok.

* As a widget I chose the same widget used to represent transactions in the home page. However, I didn't want to repeat for each transaction which account belong to (since it seemed redundant to me), so I just removed in the widget the bang operator (`!`), hence allowing for empty names in the bank account name field (commit [78f361b](https://github.com/RIP-Comm/sossoldi/commit/78f361b90bc34733fe6a1818fd6b831be79e310b))

* Finally, I added the transaction list to the account page. Moreover, I changed the color of the back arrow, since it had the same color of the background (commit [f29f92e](https://github.com/RIP-Comm/sossoldi/commit/f29f92e470cc34aa2a6846e0e8a9a9d6a3053f9a))

### DIFF
--- a/lib/custom_widgets/transactions_list.dart
+++ b/lib/custom_widgets/transactions_list.dart
@@ -63,15 +63,17 @@ class _TransactionsListState extends State<TransactionsList> with Functions {
   Widget build(BuildContext context) {
     return transactions.isNotEmpty
         ? SingleChildScrollView(
-          padding: widget.padding,
-          child: DefaultContainer(
+            padding: widget.padding,
+            child: DefaultContainer(
               child: Column(
                 children: transactions.map((transaction) {
                   int index = transactions.indexOf(transaction);
-                  bool first =
-                      index == 0 || !transaction.date.isSameDate(transactions[index - 1].date);
+                  bool first = index == 0 ||
+                      !transaction.date
+                          .isSameDate(transactions[index - 1].date);
                   bool last = index == transactions.length - 1 ||
-                      !transaction.date.isSameDate(transactions[index + 1].date);
+                      !transaction.date
+                          .isSameDate(transactions[index + 1].date);
 
                   return Column(
                     children: [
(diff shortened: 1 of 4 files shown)


theperu (approved):
LGTM, I think we can move forward and merge this

mikev-cw:
Very nice! 
Just one question about this:

> Note: Here I hardcoded that the widget will retrieve just the last 50 transactions (due to efficiency reasons). This is ugly, I admit, but for the time being it should be ok.

Since the graph is showing current month, it would make sense show all transactions of **that** month, so only those that contributes to the graph data?
I think it would be more "logical", but if you agree with this, we could handle it on another PR, and merge this meanwhile.

theperu:
@mikev-cw I would merge this as is for now, I think it makes sense in this moment to just keep it simple and then we can think about adding filters and so on when we will be ready

PR #153, opened 2024-02-12
CurrencySelector #70

K-w-e:
I managed the currency selection. 
I created a basic provider.
Then, obtaining the value through it, I display the selected currency within the application. 
Additionally, some StatelessWidget have been converted to ConsumerWidget just to read the selected currency value. 

Feel free to provide any suggestions or requests for modifications!

### DIFF
--- a/lib/providers/currency_provider.dart
+++ b/lib/providers/currency_provider.dart
@@ -1,0 +1,28 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+import '../constants/constants.dart';
+import '../model/currency.dart';
+
+final currencyStateNotifier = ChangeNotifierProvider(
+  (ref) => CurrencyState(),
+);
+
+class CurrencyState extends ChangeNotifier {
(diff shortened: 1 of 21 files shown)


lib/providers/currency_provider.dart
```diff
@@ -0,0 +1,35 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+import '../constants/constants.dart';
```
jackrua, line 4:
```suggestion

```

jackrua (commented):
Wonderful job! this currency thing it's something I had on my issue list for a while. Tried this PR on my phone and it seems that everything works as it should. 

Just to be pedantic, my Editor tells me that this is an unused import.

lib/providers/currency_provider.dart
K-w-e: Oh you are right, i will fix it!

theperu (approved):
Thanks for taking the time to do this, it was probably quite boring 😅
It seems to be working as expected and I think we can merge

PR #158, opened 2024-02-20
Feature/budget

K-w-e:
I'm sending you the pull request for the implementation of the Budgets feature, just to keep you informed about the direction I took during development. 

Some points of attention already known:

- I removed the "Add category budget" button from the planning page. Now, if you want to modify the budgets, you need to press "Manage." (I did this for both simplicity and because the flow to follow by pressing that button was not clear).
- I added the ability to delete the budget via swipe. In the UI, there was both that option and the option through a "-" button next to the budget. We need to decide which one to keep.
- When you add a new budget or modify an existing one, provide the option to select only the categories that are not already present. This should be easy to fix, but I haven't found a clean way to do it yet.

For any doubts, questions, or suggestions, feel free to reach out!

### DIFF
--- a/lib/model/budget.dart
+++ b/lib/model/budget.dart
@@ -85,6 +85,35 @@ class BudgetMethods extends SossoldiDatabase {
     return item.copy(id: id);
   }
 
+  Future<Budget> insertOrUpdate(Budget item) async {
+    final db = await database;
+
+    final exists = await checkIfExists(item);
+    if (exists) {
+      print("UPDATE ${item.toJson()}");
+      await db.rawQuery("UPDATE $budgetTable SET amountLimit = ${item.amountLimit} WHERE idCategory = ${item.idCategory}");
+    } else {
+      print("INSERT ${item.toJson()}");
+      await db.insert(budgetTable, item.toJson());
+    }
+
+    return item.copy(id: item.id);
+  }
+
+  Future<bool> checkIfExists(Budget item) async {
+    final db = await database;
+
+    try {
+      final exists = await db.rawQuery("SELECT * FROM ${budgetTable} WHERE ${item.idCategory} = idCategory");
+      if(exists.isNotEmpty) {
+        return true;
+      }
+      return false;
+    } catch (e) {
+      return false;
+    }
+  }
+
   Future<Budget> selectById(int id) async {
     final db = await database;
 
@@ -134,5 +163,17 @@ class BudgetMethods extends SossoldiDatabase {
     final db = await database;
 
     return await db.delete(budgetTable, where: '${BudgetFields.id} = ?', whereArgs: [id]);
+  }
+
+  Future<int> deleteByCategory(int id) async {
+    final db = await database;
+
+    return await db.delete(budgetTable, where: '${BudgetFields.idCategory} = ?', whereArgs: [id]);
+  }
+
+  //PER TEST
+  Future<int> deleteAll() async {
+    final db = await database;
+    return await db.rawDelete("DELETE FROM $budgetTable");
   }
 }
--- a/lib/model/budget.dart
+++ b/lib/model/budget.dart
@@ -170,10 +170,4 @@ class BudgetMethods extends SossoldiDatabase {
 
     return await db.delete(budgetTable, where: '${BudgetFields.idCategory} = ?', whereArgs: [id]);
   }
-
-  //PER TEST
-  Future<int> deleteAll() async {
-    final db = await database;
-    return await db.rawDelete("DELETE FROM $budgetTable");
-  }
 }
--- a/lib/pages/planning_page/manage_budget_page.dart
+++ b/lib/pages/planning_page/manage_budget_page.dart
@@ -1,0 +1,145 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:sossoldi/providers/budgets_provider.dart';
+
+import '../../model/budget.dart';
+import '../../model/category_transaction.dart';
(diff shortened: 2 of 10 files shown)


lib/model/budget.dart
```diff
@@ -135,4 +164,16 @@ class BudgetMethods extends SossoldiDatabase {
...
+    final db = await database;
+
+    return await db.delete(budgetTable, where: '${BudgetFields.idCategory} = ?', whereArgs: [id]);
+  }
+
+  //PER TEST
+  Future<int> deleteAll() async {
+    final db = await database;
+    return await db.rawDelete("DELETE FROM $budgetTable");
+  }
```
mikev-cw, line 178: Remember to remove this on the final submission, if is a method that you just need to simplify your work and it won't go in the final release

lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -0,0 +1,161 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:sossoldi/providers/budgets_provider.dart';
```
mikev-cw, line 3: don't use `package:sossoldi` for our own packages, just use the absolute path

lib/pages/planning_page/planning_page.dart
```diff
@@ -1,4 +1,6 @@
 import 'package:flutter/material.dart';
+import 'package:sossoldi/model/budget.dart';
```
mikev-cw, line 2: don't use `package:sossoldi` for our own packages, just use the absolute path

lib/pages/planning_page/widget/budget_card.dart
```diff
@@ -1,13 +1,18 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:sossoldi/pages/planning_page/manage_budget_page.dart';
+import 'package:sossoldi/providers/transactions_provider.dart';
```
mikev-cw, line 4: don't use `package:sossoldi` for our own packages, just use the absolute path

lib/pages/planning_page/widget/budget_category_selector.dart
```diff
@@ -0,0 +1,112 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:sossoldi/model/budget.dart';
+import 'package:sossoldi/model/category_transaction.dart';
```
mikev-cw, line 5: don't use `package:sossoldi` for our own packages, just use the absolute path

mikev-cw (commented):
Ok, that's very cool!!!
I just made some minor notes, but I've nothing to say about the overall approach.
Also keep an eye to those brand new conflicts.

About this: 

> I added the ability to delete the budget via swipe. In the UI, there was both that option and the option through a "-" button next to the budget. We need to decide which one to keep.

It seems that in the design file @federicopozzato has planned kind of "edit mode", where user can tap the "-" button to delete a budget, but since you can in fact edit all about budgets when you press "Manage", actually having an extra edit mode can be a bit redundant at this stage?

federicopozzato:
@mikev-cw @K-w-e thanks for pointing out those things!
The redundancy of the actions has been done on purpose, to give the user more freedom, even though I understand what you say about too many actions, and you may be right.
For semplicity let's keep it in the way you did, it's fine!

lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -0,0 +1,162 @@
...
+        const SizedBox(height: 10),
+        Text(
+            "Your monthly budget will be: ${budgets.isEmpty ? 0 : budgets.fold(0, (sum, e) => sum + e.amountLimit.toInt())}€"),
+        const SizedBox(height: 10),
+        const Divider(height: 1, color: grey2),
+        Padding(
+            padding: const EdgeInsets.all(8),
+            child: SizedBox(
+                width: double.infinity,
+                child: ElevatedButton(
```
theperu, line 144: Could you make the button style similar to the other ones in the app? For example the button to add/edit a transaction.

lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -0,0 +1,162 @@
...
+                ),
+                Text(
+                  'Amount',
+                  textAlign: TextAlign.right,
+                ),
+              ],
+            )),
+        Expanded(
+            child: Column(children: [
+          ListView.builder(
```
theperu, line 75: Below the list of categories we should have a text that explains to the user that if they want to delete a budget they should swipe it left or right.

theperu (changes requested):
I really like the implementation, good job! I left a couple of comments for minor changes but apart from that it seems fine to me.
Since you have already done the work on the budgets could you also work on the graphs that are at the bottom of the Dashboard next? 

FYI I just noticed that there are also some conflicts that need to be resolved before we can merge

K-w-e:
PR updated!
That file with conflicts is caused by the fact that that widget has been rearranged and almost completely modified, I don't think I can resolve it myself. You should accept the changes from my branch through the PR.

Let me know if i can help

@theperu

lib/pages/planning_page/manage_budget_page.dart
K-w-e: Fixed

lib/pages/planning_page/manage_budget_page.dart
K-w-e: Added

PR #161, opened 2024-03-26
Update Privacy Policy

theperu:
I updated the Privacy Policy section to something that should reflect more what we are currently doing. Also it should be enough to pass the review from the stores once we submit the apps.
@mikev-cw and @lucaantonelli I need confirmation that we do not currently gather any type of informations from our users like analytics or stuff like that.


mikev-cw (approved):

lucaantonelli:
LGTM, we can merge it.

PR #162, opened 2024-03-29
onboarding and minor layout fixes

fres-sudo:
the minor layout fixes regards:
  - month selector
  - budgets provider (create another method "insertOrUpdates" in the budgets provider is also possible)
  - add category page
  - routes


theperu:
Nice work! Most of the things look good to me, I just noticed a couple of issue that need to be fixed before merging:
1. When I close and reopen the app I see the onboarding again. Once the user completes the Onboarding the first time they should skip to the dashboard.
2. When I add the main account I can only input a number without the decimal points (Ex. I can add 3000 but not 3000.01)

Here is a video:
https://drive.google.com/file/d/1kiv-2eVfrE10KS8I2Y6FAEKEElZP4ypB/view?usp=drive_link

fres-sudo:
Damn, I implement the one-time onboarding feature but for testing purpouse i commented it.
The idea was to uncomment it but i just forgot before committing >.<.
As you can see in main.dart at line 47/48 ther's the correct line and a TODO that explains everythings.

For the decimal point i'll take a look and see what i can do.

Thanks!

lib/pages/onboarding_page/widgets/add_budget.dart
```diff
@@ -0,0 +1,88 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:sossoldi/constants/functions.dart';
```
lucaantonelli, line 3: You use a mix of relative and absolute path import, we try to use relative import across the project, just an easy fix, i noticed that also in routes.dart and some other widgets :)

theperu (approved):
@lucaantonelli now it seems fine to me, if you agree we can merge 🙂

PR #165, opened 2024-04-13
Fixed some UI issues related to dark mode

diegodiome:
- Patched #124
- Fixed some issues


theperu:
Hi! I noticed that the planning page looks a bit strange and I cannot see what I am doing when I manage them. Also in the transaction page there are some text in dark blue that I think are quite hard to read, maybe we could switch them to the same grey that we use for "Add a description".
Apart from these two things I think it looks good, thanks!

theperu (changes requested):
I see that you solved most of the issues, unfortunately now the onboarding doesn't look great if you open the app and you have the dark mode set as default. If it's not an easy fix maybe we could ignore the theme for the onboarding and check it only after it.
Also in the planning page when I tap on manage in the top right the modal looks a bit off and I cannot read the texts for each budget set

Issue #49, opened 2023-01-07
Graphs Page 


theperu:
We need to fill the Graphs page that is currently empty. The best thing I think it is to divide it in three task, one for each widget. Here you can find the [Figma](https://www.figma.com/file/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?node-id=976%3A2601&t=0LDo9rqpYWzMm6MJ-0)

- [x] Graph
- [x] Accounts
- [x] Categories

theperu:
Since I have done a little bit of work with the line charts I'll take the upper part of the page. If there are other people that want to take on the rest of the page feel free to comment here

napitek:
I would like to work on it

theperu:
Great! I'll assign this to you then

PR #166, opened 2024-04-21
CategoryTransactionType for CategoryTransaction model, Charts in GraphsPage

napitek:
### Description

This PR can be divided into two main points:

- CategoryTransactionType for categories:
  until now the income and expense type was only for transactions. It makes sense to have the type for categories as well.
  
  _the **"CategoryTransaction"** model has changed, you will need to reset the DB before proceeding_
  
- GraphsPage refactoring with addition of charts in reference to the issue #49 

### Notes
The BarChart is not quite as functional as I would like. i put it in and you can click on the bars to change the month and update the category data.
I wasn't too convinced about being able to change the month at the bottom versus the data updating at the top. 
In the future we could integrate the month_selector.


theperu (changes requested):
Hi! Thanks for all the work that you have done. I found one issue, when I tap on a bar month in which I have zero euro spent (ex. I tap on May which is in the future) then the widget updates but I can't see the bars anymore. The only way to restore it is if I close and reopen the app. Let me know if you are able to reproduce this and/or you need a screen recording.
I think that there are some conflicts that need to be resolved before merging

napitek:
Hi @theperu. Yes I just looked and I noticed what you say. 
I'll fix it and commit it.
Although as said before, that graph needs to be reviewed and discussed for a moment.

napitek:
the GraphsPage needs a month_selector, which cannot be exclusively the final chart.
Since one already exists, I would put that in for now.

theperu (changes requested):
I finally found the time to look at your changes. I notice another two things that might need an improvement. If a category as no transaction in the current month I see NaN instead of 0, when this happens I also see the bars for the previous months to 0 even if there are transactions on it. You can easily reproduce this by having only one income category with multiple transactions in the previous months but not in the current one

EDIT: Also the month bars are all the same height in each month

napitek:
> I finally found the time to look at your changes. I notice another two things that might need an improvement. If a category as no transaction in the current month I see NaN instead of 0, when this happens I also see the bars for the previous months to 0 even if there are transactions on it. You can easily reproduce this by having only one income category with multiple transactions in the previous months but not in the current one

I didn't notice this bug. I'll take care of it in these days


> EDIT: Also the month bars are all the same height in each month

the bars are all the same because the graph does not yet calculate the "average". 
I don't think there is even a method in general that does that (maybe I'm wrong and haven't seen it). 
It will have to be implemented.

theperu (approved):
LGTM! I'll wait for a review from either @lucaantonelli or @mikev-cw before merging

mikev-cw (approved):
Yeah, it's good, well done!

There's just one behavior a bit tricky on graphs page: when you change the month the page scrolls to the top, creating an unpleasant effect for the user.
But I think we can merge this meanwhile, and fix on next contributions

lucaantonelli:
LGTM apart from the problem reported by Mike, i 'm merging it.

PR #170, opened 2024-07-21
Google Drive support + Bug fixing + Web app fixes

matteosz:
# Google Drive support + Bug fixing + Web app fixes

This PR introduces many improvements to the project, mainly a big new feature and several bug fixes.

### Why?
The need of a cloud storage for the application is critical: first, users want to use the application from multiple devices (e.g. adding transactions on fly from the smartphone, but then analysing the net worth from a laptop). Second, there's an evident big reliability problem (SPOF) in storing everything locally. Thus, I consider this feature of a much higher priority than the one provided in the development plan.

### How?
The user goes to the settings, and find the Google Drive option, from where they can authenticate with their Google Drive account. Since that moment, automatic sync will be activated. It will work in the following way: 
1. The user logs in with Google Drive for the first time: immediately a pull from Google Drive is executed, and if the database is present it would be downloaded and overwritten to the current one.
2. The user has the possibility, always through the settings, to force a push/pull to/from Google Drive at any time, besides the capabilities of the automatic sync
3. Every time the user closes the app, the database is pushed to Google Drive and persisted (updating the old database if already present or creating a new one if absent)
4. Every time the user opens the app, given that they previously logged in to Google Drive and never logged out, the database is pulled and overwritten to the local one
5. The user can disconnect easily the Google Drive connection from the settings, and the sync will stop

Here's a quick recording of the functionality in action:

https://github.com/user-attachments/assets/17015e5c-e409-4c13-b559-5c0aa43c6d03

_Current limitation_
This feature currently works for Android, iOS and MacOS. Windows and Linux still don't have support for the google_sign_in plugin. I implemented the web application support for the google authentication; however, there still remains the challenge of downloading/overwriting/uploading the local database due to the constraints of accessing local storage typical of web apps (see TODOs in the code).

### Previous issues
1. The web application wasn't working at all, mostly due to some wrong usage of dart.io Platform object, which is not available on web. I've replaced all these occurences with conditional imports, ensuring it's not web before checking the platform.
2. The local storage (sqlite3) wasn't working for web. By using sqflite_ffi_web I managed to have it working without the need of changing database type only for web (only needed to change the factory method)

### Current limitation
There are still important things not working on the web version (setting the date of a transaction, broken search), but overall the app became at least usable now, and further bug fixes can be done in future PRs.

## Bug fixing: what and how?
The following bugs have been addressed and fixed:
1. Inserting an expense value wasn't working, this was due to the sign '-' appended at the beginning and following checks; this was causing an user typing the second digit to reset the previous value inserted (impossible to insert any value with more than 1 digit). For simplicity, I just removed the minus sign, as the red color is enough to understand it's an expense (and income doesn't have a plus sign, so it's more symmetrical).
2. Fixed many linter errors and warnings, by running the automatic dart tool to solve them
3. When a new transaction is added, the bank accounts' values don't change accordingly. I ensured to refresh the provider for the accounts when a new transaction is added. 
-> **TODOs**: I noticed this happens also with budgets, they are not updated in real time, or when transactions are edited/removed.
4. Solved some build issues for MacOS


mikev-cw:
Hello, that's impressive, thank you!
Just one speedy question before going deeper: as far as i know, when you need to use Google's API or services, you need to create an app on developers profile, that collect scopes and authorizations of users. Did you register an application with your account then?

mikev-cw:
I'm closing this as decided at our last meeting. Head on Discord if you want to reach us.
Thanks!

PR #173, opened 2024-08-15
Fix for Account Exclusion Issue Due to Recurring Transactions

gioisco:
This pull request addresses a bug where active bank accounts with only recurring transactions were being excluded from the query results.

### Changes:
- Corrected the SQL query to ensure that active bank accounts with only recurring transactions are no longer excluded from the results.
- Added a new `recurringFilter` variable to clearly define and encapsulate the logic for filtering out non-recurring transactions.
- Adjusted the `LEFT JOIN` clause to apply the recurring filter within the `ON` condition, preserving all active accounts while accurately calculating their balances.
- This fix resolves an issue where certain accounts disappeared from the query results, ensuring all active accounts are correctly handled in the balance calculation.

### Issue Demonstration:

This video demonstrates the issue that this pull request resolves:

https://github.com/user-attachments/assets/ff05cc92-8d7b-4b3e-871d-bd51d4267d42

### Testing:

- Verified that all active accounts are now included in the query results, even those with only recurring transactions.
- Conducted extensive testing to ensure that the fix does not introduce any regressions and that the balance calculations remain accurate.

Please review the changes and let me know if you have any feedback. Thank you!

### DIFF
--- a/lib/model/bank_account.dart
+++ b/lib/model/bank_account.dart
@@ -149,8 +149,8 @@ class BankAccountMethods extends SossoldiDatabase {
     final db = await database;
 
     final orderByASC = '${BankAccountFields.createdAt} ASC';
-    final where =
-        '${BankAccountFields.active} = 1 AND (${TransactionFields.recurring} = 0 OR ${TransactionFields.recurring} is NULL)';
+    final where = '${BankAccountFields.active} = 1 ';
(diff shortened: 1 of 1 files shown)


mikev-cw:
Nice catch! I confirm the bug, and I confirm this fixes it.
Let's merge

Issue #178, opened 2024-11-07
Hide subcategory button


mikev-cw:
We should temporarily hide the subcategory button on edit category page. This will we implemented on future releases.
Since we added the "type" option, it also hides the "delete" button (the page is not scrollable)

![Screenshot 2024-11-07 alle 09 41 23](https://github.com/user-attachments/assets/efe462f9-539c-452e-9fe7-9d6583029b73)

Issue #176, opened 2024-11-07
Hide developers bar


mikev-cw:
Before releasing our MVP, we should hide the developers bar on bottom of the settings page. 
This bar should be activated by the user somehow, for instance:

A double tap on the app version?
A dedicated option inside "General Settings"?

You choose.
Whatever the case, the user should be warned that the database and its data will be deleted forever. 

![Screenshot 2024-11-07 alle 08 58 49](https://github.com/user-attachments/assets/0aab14fb-1885-467c-8a86-b39111b2a85a)

PR #179, opened 2024-11-07
Hide dev bar and subcat button

mikev-cw:
This aims to address Issues #178 and #176
To toggle the dev bar, you need to tap 4 times the word "Settings" on top of the settings page


theperu (approved):
Looks good, I noticed that we need to re-enable the dev bar every time we leave the settings but it's not an issue for what it needs to do now 👍🏻

PR #184, opened 2024-12-05
Bug fix: Prevent user to create budgets for income categories

theperu:
Currently a user could create an income category and then add a budget to it. This PR aims at resolving this by doing the following things:
1. During the Onboarding phase the user can only create Expense categories and add a budget to them
2. In the planning page when I user wants to create a new budget only the Expense categories are shown


mikev-cw:
Good catch!

mikev-cw (approved):

Issue #201, opened 2025-01-28
Bug: Account name can't be more than 9 letters in the Onboarding


theperu:
# 🐛 Bug Report

## 📋 Description

**Summary**:  
A user is not able to create an account with more than 9 letters and no numbers

**Expected Behavior**:  
The user should be able to create an account with more than 9 letters

**Actual Behavior**:  
After typing more than 9 letters the text field resets.
---

## 🧪 Steps to Reproduce

1. Start the Onboarding
2. In the page for creating your first account add a name with more than 9 letters
3. The text field will resets
4. Add as many numbers you want
5. The text field doesn't resets

lucaantonelli:
Fixed with #207

PR #207, opened 2025-02-15
Fix #194 & #201

### LINKED ISSUE #194: Remove transfers from and to same account
Android app

lucaantonelli:
- Solve https://github.com/RIP-Comm/sossoldi/issues/194 (it was possible to reproduce selecting the toAccount, then in the fromAccount was possible to select the same, now disabled)
- Solve https://github.com/RIP-Comm/sossoldi/issues/201 removed unnecessary validation
- Improve and simplify some providers for creating/updating both categories and accounts, removed when unnecessary switching to simple setState for basic usages.
- Improve more info and collaborators pages
- Improve the onboarding page layout to match the figma design


theperu (changes requested):
Great job! 
I just noticed an issue on the Onboarding while testing on Android (I can't test iOS at the moment). When I try to add a category I get a failed assertion for the dropdown, this is probably because in `add_category.dart` we have `CategoryTransactionType categoryType = CategoryTransactionType.income` which defaults the value for the new category to income while this is not allowed in the Onboarding (see the pr #184).
Apart from this I haven't found other issues and I will add a more in depth review after your changes.

lucaantonelli:
> Great job! I just noticed an issue on the Onboarding while testing on Android (I can't test iOS at the moment). When I try to add a category I get a failed assertion for the dropdown, this is probably because in `add_category.dart` we have `CategoryTransactionType categoryType = CategoryTransactionType.income` which defaults the value for the new category to income while this is not allowed in the Onboarding (see the pr #184). Apart from this I haven't found other issues and I will add a more in depth review after your changes.
Should be fixed now.

theperu (approved):
LGTM! The error is fixed and I haven't found new issues 👍🏻

I just have a few points that we should keep in mind for future PRs:
1. We should create a separate PR when we want to upgrade Flutter, this will make it easier to check if we have issues that are specific to it
2. We need to agree on a linter that would help format the code in the same way for everyone. PRs like this one are harder to review because many changes are just for formatting changes.
3. We should try to have each commit fix a specific thing so that we can review them one by one instead of having to test multiple stuff at once. Of course this is more for complex changes in which we have to touch multiple things and not trivial ones

Issue #200, opened 2025-01-28
Bug: No confirmation message is shown when deleting an account


theperu:
# 🐛 Bug Report

## 📋 Description

**Summary**:  
A user doesn't see a confirmation message when deleting an account.

**Expected Behavior**:  
The user should be able to cancel the deletion or confirm

**Actual Behavior**:  
The account gets deleted and the user has no way to revert the action
---

## 🧪 Steps to Reproduce

1. Create an account
2. Go to Settings -> Accounts
3. Select the account
4. Delete the account
5. No confirmation message is shown and the account is not there anymore

mikev-cw:
I think we need a confirmation prompt for transactions too

theperu:
Yes, I agree. Every action that is going to delete something should have a confirmation dialog first

diegocurb:
Hey, this issue would be my first open source contribution on Github. Would be happy to resolve it if help is needed! I could help come up with the confirmation dialog and create the pop up

mikev-cw:
@diegocurb Thank you, we appreciate your help! I'm gonna assign this to you, feel free to reach us and ask anything you need. Good work

bongio94:
I'm working on this now 🙂

bongio94:
@theperu , this is pretty much done.

I'd wait for #239 to be completed before opening a new one because I mistakenly pushed on my forked `main` so the branch which contains this fix is ahead of the repo's `main`.

PR #208, opened 2025-02-16
First Open-Source Contribution: Fixing Bug #200

amandajiang259:
Hello!

I’m excited to be making my first open-source contribution! I noticed the “good first issue” labeled under _#200: Bug: No confirmation message is shown when deleting an account_, and I decided to take a look. I ran the code on my end, and it appears to be working correctly now. A confirmation message is displayed, allowing the user to either delete their account or cancel the deletion process.

Thank you for your time, and I hope you’ll consider my fixes!

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -4,6 +4,8 @@ import 'package:intl/date_symbol_data_local.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 import 'package:timezone/data/latest.dart' as tz;
 import 'package:timezone/timezone.dart' as tz;
+import 'package:sqflite_common_ffi/sqflite_ffi.dart';
+import 'package:path_provider/path_provider.dart';
 
 import 'model/recurring_transaction.dart';
 import 'providers/theme_provider.dart';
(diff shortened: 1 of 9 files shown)


amandajiang259:
> Hi! Thanks for your contribution and welcome to the project! Before going ahead with the full review I noticed that you added a commit for some Windows related stuff, since this is not specific to the bug #200 I think it would be better to revert those and only keep the ones about the confirmation dialog. If you want you can then open a different PR for making the program run on Windows in which you explain better the issue that it is solving. This would facilitate a lot the review process 😁

Okay, got it. Thanks! I'll work on that.

theperu:
@amandajiang259  do you have any news on this? Otherwise I'll have to close the PR and re-assign it 😅

bongio94:
Hey,

I can take this task if the original contributor can't continue his work.

I can see an `AlertDialog` has been used in the proposed solution. 

What is the approach to dialogs in the app? Do you plan to have a native look and feel for these kind of UI elements (I can see you're using the adaptive Switch so at first glance it would make sense to use the adaptive AlertDialog as well)?

theperu:
I'll close this one, @bongio94 I can assign the issue to you and you can open a new PR when it's ready. Using an adaptive AlertDialog would probably be the best approach 👍🏻

PR #210, opened 2025-02-19
Fix #202 - #203 + Dark Mode improvements

### LINKED ISSUE #157: Some issues with Dark Mode
If you set Dark Mode, the colors in the line chart do not change. 
This point is already addressed in the code: 

https://github.com/RIP-Comm/sossoldi/blob/fc3c316695686c0556543fac5da24b7b8b2b6676/lib/pages/home_page.dart#L139-L145

We could use something like `blue6` (`0xFF7DA1C4`) to get a nice contrast: 

p3

Moreover all the account and category icons should be switched to Dark Mode in different pages (only in the settings page they are updated while, for example, in the home page they are not). 

Finally in the transactions page, if you switch to Dark Mode the month selector creates so much contrast that is almost unreadable: 

p4

### LINKED ISSUE #202: Bug: Dark mode resets after restarting the app
# 🐛 Bug Report

## 📋 Description

**Summary**:  
Dark mode resets after restarting the app

**Expected Behavior**:  
If the user sets the dark mode then it should be persistent even after closing and reopening the app

**Actual Behavior**:  
After setting the dark mode, when a user restarts the app he sees the light mode again
---

## 🧪 Steps to Reproduce

1. Go to Settings -> General Settings
2. Set the Dark mode
3. Close and reopen the app
4. The Dark mode is off again

### LINKED ISSUE #203: missing decimals in new account prompt
# 🐛 Bug Report

## 📋 Description

**Summary**:  
When creating a new account, iOS users cannot add decimals for starting values

**Expected Behavior**:  
The user should be able to add decimals for accounts starting values

**Actual Behavior**:  
Keyboard type does not have decimal point

Image

lucaantonelli:
Fix #202 - Solved in theme_provider.dart
Fix #203 - Solved in add_account.dart (changed only the regex used in the text field)
Fix #157 - Overall improvements of dark mode across app, there is consistency now between pages.

Commented out release config in build.gradle for github pipelines to work.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -4,11 +4,11 @@ plugins {
     id "dev.flutter.flutter-gradle-plugin"
 }
 
-def keystoreProperties = new Properties()
-def keystorePropertiesFile = rootProject.file('key.properties')
-if (keystorePropertiesFile.exists()) {
-    keystoreProperties.load(new FileInputStream(keystorePropertiesFile))
-}
+// def keystoreProperties = new Properties()
+// def keystorePropertiesFile = rootProject.file('key.properties')
+// if (keystorePropertiesFile.exists()) {
+//     keystoreProperties.load(new FileInputStream(keystorePropertiesFile))
+// }
 
 android {
     namespace "com.ripster.sossoldi"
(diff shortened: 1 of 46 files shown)


theperu (changes requested):
Good job! I just noticed that there is still a couple of section that might be improved while using the dark mode:
1. The list of transaction for a specific category in Transactions -> Categories/Accounts -> Expand one of the categories ([Screenshot 📷](https://github.com/user-attachments/assets/43edcea6-b663-4046-844a-960e87ae094b))

2. The list of budgets in Planning -> Manage ([Screenshot 📷](https://github.com/user-attachments/assets/f342d302-b64e-4843-ac7f-9a7859481d32))

theperu (approved):
Thanks, LGTM! I'll proceed to merge this 👍🏻

PR #224, opened 2025-02-23
Fix #211 - #215 - #217 + New custom widget: RoundedIcon

### LINKED ISSUE #211: Space between icon and text in account widget.
In the dashboard, in the account element there's no enough space between the icon ant the text

Image

### LINKED ISSUE #215: iOS: Not enough space between “See older payments” and the above details
iOS: Not enough space between “See older payments” and the above details. Please check better all the spaces and paddings in Figma

Image

### LINKED ISSUE #217: iOS: Switcher is different between "recurring payment" pages
iOS: If I tap “add recurring payment” the “Recurring payment” switcher in the transaction panel is different than the one that i see when I’m in the transaction page coming from the plus icon. It also looks disabled

Image

Image

lucaantonelli:
This PR fixes current issues:
- Fix #211 (just added spacing)
- Fix #215 (just added spacing)
- Fix #217 (made the switch adaptive to platform)
- I also implemented a new custom widget called RoundeIcon, since we use a lot this type of icons across the app, i made it resuable, so we also have less code in every widget that uses that component

### DIFF
--- a/lib/custom_widgets/accounts_sum.dart
+++ b/lib/custom_widgets/accounts_sum.dart
@@ -7,14 +7,15 @@ import '../constants/functions.dart';
 import '../constants/style.dart';
 import '../providers/accounts_provider.dart';
 import '../providers/currency_provider.dart';
+import 'rounded_icon.dart';
 
 /// This class shows account summaries in the dashboard
 class AccountsSum extends ConsumerWidget with Functions {
   final BankAccount account;
 
   const AccountsSum({
-    super.key,
     required this.account,
+    super.key,
   });
 
   @override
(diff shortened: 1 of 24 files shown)


theperu (changes requested):
Thanks! About #211 it seems to me that there is still not enough space, apart from this everything else looks good 👍🏻
Screenshot 2025-02-23 122511

lucaantonelli:
> Thanks! About #211 it seems to me that there is still not enough space, apart from this everything else looks good 👍🏻 Screenshot 2025-02-23 122511

Can you double check? Because this is how it shows on my end:
photo_2025-02-23_12-32-52

theperu (approved):
@lucaantonelli you are right, I re-installed the app and now it looks fine. Not sure what happened there 😅. All good then!

PR #225, opened 2025-02-23
Import export feature

theperu:
Adding Import/Export feature

**IMPORTANT NOTE FOR THE REVIEWER:** currently I'm not able to test on iOS which means that this code was only tested with a few Android devices. I looked online for what the right approach might be but it needs to be verified. Note also that this PR adds new request permissions which we might need to update in the stores

## 🎯 Purpose

Adding the possibility for our users to manually backup their data with a CSV file and import it back 
## 📱 Changes Made
Describe the key changes implemented in this PR:
- **UI Changes**: Added a Import/Export section in the settings
- **Dependencies**: 
      - flutter_phoenix: ^1.1.1 (for the app restart after the import)
      -  csv: ^6.0.0 (to handle the csv easily)
      -  path_provider: ^2.1.5 (to navigate the folders)
      -  file_picker: ^8.1.7 (to select the file)
      -  device_info_plus: ^11.2.0 (to handle the permissions requests on Android)

## 🧪 Testing Instructions
1. Go to settings and try exporting the data
2. Clear the DB
3. Import the data back

## 📸 Screenshots/Videos
[Video demonstration of the feature 📽️](https://drive.google.com/file/d/1Yz5_jbMca_15OCdAtskPQ7eyz-ARw8O3/view?usp=sharing)

## 🔍 Review Checklist
- [ ] Code follows project style guidelines
- [ ] Documentation updated
- [ ] Widgets are properly disposed


mikev-cw:
Tested on iOS (simulator 18.1, phisical 18.3).
The process of importing/exporting is working very well overall!
I noticed a weird behavior however, on both devices: Every time i tap on "Export Data", a file called `exported_database.csv` is created automatically on the default path (local files > Sossoldi > sossoldi_exports), plus an extra file called `sossoldi_export_[digits].csv` (that seems to be a clone of the first file), without any confirmation my side. 
Then, if i cancel the export and tap again on "Export Data", another file called `sossoldi_export_[digits].csv` is created without confirmation. And so on, every time i tap on the export button (see attached video).
Is that supposed to be the correct behavior?


https://drive.google.com/file/d/11q0oV-idZGMzLVn5AV4-9yIxc8EuFtcS/view?usp=sharing

theperu:
@mikev-cw I made some changes that hopefully will fix the weird behavior that you were seeing before, now it should only save one file called `sossoldi_export_[timestamp].csv`. I think the issue was caused by the fact that I was saving the file twice and I didn't notice since on Android I was just seeing one.
The expected behavior should be that when you go to export the data it should let you selected a folder and then save it there, if the export is aborted then no file should be saved.

Hopefully now it works fine, if there are other issues related to iOS I might need the help of one of you since I don't have a way to properly develop and test on it (for now, I'll probably get one soon 😅)

lucaantonelli:
Tested on both ios and Android, LGTM.

lucaantonelli (approved):

Issue #228, opened 2025-02-25
Currency not created if not finisehd onboarding


lucaantonelli:
Currency should be created within the db since are static, and not during the onboarding. If new currency are going to be supported in future, then we should add a migration for them.

Issue #226, opened 2025-02-23
Add New Category show last selected


lucaantonelli:
Steps to reproduce:
- Go to Settings > Category > Select one category
- Then go back to home page and try to add a transaction
- Open the categories dialog
- Pressing then new category button show information about previously selected category, should be blank instead

Same problem occurre also for the accounts.

PR #231, opened 2025-03-01
Fix #221 + #226 + #228 + Minor fixes

### LINKED ISSUE #221: iOS: Check the font across all the app
I noticed that the font is not correct in the account widget in the dashboard. (you can see below the difference with the Figma)

The font across the whole app is NUNITO SANS, it's a Google font. So please while fixing the font in this widget, check it across the whole app.

Image

Image

lucaantonelli:
This PR introduces this list of fix/improvements:
- Fix Colors of text in textfield for darkmode
- Add correct google font from figma
- Moved currency creation on database setup instead of onboarding
- Show updated account amount when adding/editing a transaction
- Fix for notification on android release and enabled also for ios


theperu (changes requested):
Great work! Could you update the budgets in the Dashboard as you already did with the accounts? 
After that we can probably merge and release a new version 👍🏻

lucaantonelli:
> Great work! Could you update the budgets in the Dashboard as you already did with the accounts? After that we can probably merge and release a new version 👍🏻

Added a commit about that.

theperu (approved):
LGTM! I'll proceed to merge this, I think we gathered enough changes to justify a new release in the stores 👍🏻

PR #233, opened 2025-03-03
feat(account, transaction): reconciliation and transaction list

### LINKED ISSUE #197: Feature: Adjust balance of the account
# 🙋🏻‍♂️ Adjust balance of the account

## 📋 Description

**Summary**:  
Currently a user can only update the balance of the account with a transaction. We should implement a way to change it manually

**Expected Behavior**:  
From the settings or the account page a user should be able to set the new balance of the account

**Notes**:  
An easy way to implement this could be a special transaction that adds/removes the expected account and it is called something like Reconciliation

fres-sudo:
This PR aims to add the possibility to change the total amount of your bank account. 

Steps to take a reconciliation:

- Go to Settings Page
- Go to Accounts
- Just change the total amount and click "Update Balance"

By doing those steps, the total amount of your accounts updated and a new transaction "Reconciliation" is being created.

This PR also improve the `transaction_list.dart` code by using `ListView.separated`and `ListTile` to stick to material widget.

-Closes #197

### DIFF
--- a/lib/custom_widgets/transactions_list.dart
+++ b/lib/custom_widgets/transactions_list.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:intl/intl.dart';
 
 import '../constants/constants.dart';
 import '../constants/functions.dart';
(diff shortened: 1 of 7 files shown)


theperu (changes requested):
Great job! I had a brief look at your changes and it's working well. I would add this feature also in `account_page.dart` for the UI you could do something similar to these screenshots. After this I would also add a `Your transactions` text before the transaction list to divide more the sections.

Feel free to change the text or how it looks, it's here just to give you an idea 😁

### Before tap
Screenshot 2025-03-04 211255

### After tap
Screenshot 2025-03-04 211314

mikev-cw:
Hello, and thank you for this contribution!! I see that overall the process is working very well!
Just noticed something about the number formatting. Tested only on Android, started from 0 and filled DB with demo data.

When i tap the account, i see the current balance with a lot of decimal digits:
Screenshot 2025-03-04 alle 22 50 51

Also the reconciliation transaction is created with the same problem on the DB:
Screenshot 2025-03-04 alle 22 51 08

I think that's just the limitation of the binary representation of floating point numbers. Can you check on that? Thank you! Other than that i think that your PR is ready for the next steps!

fres-sudo:
Some updates, sorry for the delay!
@theperu I managed to implement features that you requested, lemme know if the current state seems good to you.

@mikev-cw TBH I didn't manage to find a way to replicate the issue that you are referring to (currently running on iOS), can you please give me more detailed steps to reproduce?

theperu:
Thanks @fres-sudo tomorrow I'll review your changes and give you more details about the issue reported by @mikev-cw

theperu (approved):
LGTM! 👍🏻

I tried both the update from settings and the account page and it seems to be working fine.

About the issue pointed out by @mikev-cw I am able to reproduce it only when I use the demo data and it might be related to how we create that data. I saw that those values in the DB are stored as `REAL NOT NULL` which might have some floating-point precision issue from time to time. IMHO we can proceed to merge this and explore a bit more how we generate that data.

I'll wait for Mike's opinion and then we'll merge it

PR #239, opened 2025-03-12
Resolve #219: Add input formatter to allow one period, changes to keyboardType to show numpad on iOS

### LINKED ISSUE #219: Keyboard with numbers in the transaction amount field
When I tap in the amount field on adding a new transactions, I expect to have the number keyboard, with the decimal (that should be a comma, not a point, but for this there's already another issue #112 )

Image

Image

bongio94:
In this PR:

- b8f59020541a7499bfedf915b0c5a881b18101e3: 
    - Updated `keyboardType` to have the `signed` property `true` only when running on Android. This allows to show the numpad on iOS as well.
    - Added a new `FilteringTextInputFormatter` which allows for one period only to be entered in the `TextField`.
- 58b1501575e4fb02d9854720567273636870542a: Added two tests for the `AmountWidget` widget to check that only single period are allowed in the `TextField`.

Resolve #219

### DIFF
--- a/lib/pages/add_page/widgets/amount_widget.dart
+++ b/lib/pages/add_page/widgets/amount_widget.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import "package:flutter_riverpod/flutter_riverpod.dart";
@@ -41,10 +42,24 @@ class _AmountWidgetState extends ConsumerState<AmountWidget> with Functions {
                 ),
               ),
         ),
-        keyboardType:
-            const TextInputType.numberWithOptions(decimal: true, signed: true),
+        keyboardType: TextInputType.numberWithOptions(
+          decimal: true,
+          // Leaving the default behaviour on Android which seems to be working as expeceted.
+          signed: defaultTargetPlatform == TargetPlatform.android,
+        ),
         inputFormatters: [
-          FilteringTextInputFormatter.allow(RegExp(r'\d*\.?\d{0,2}')),
+          FilteringTextInputFormatter.allow(
+            RegExp(r'\d*\.?\d{0,2}'),
+          ),
+          // AI generated regex: checks that the String contains one '.' only.
+          //
+          // Because replacementString is not nullable and defaults to empty String
+          // I've set it to the current input text so that every time a second period
+          // is inserted, it won't clear the text that's already present.
+          FilteringTextInputFormatter.allow(
+            RegExp(r'^[^.]*\.?[^.]*$'),
+            replacementString: widget.amountController.text,
+          ),
         ],
         autofocus: false,
         textAlign: TextAlign.center,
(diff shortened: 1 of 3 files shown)


theperu:
Thanks for the contribution! Assigning this to @mikev-cw since I can't currently test on iOS

mikev-cw:
Tested on iOS, on my phisical device with English locale, and a simulator in italian locale. On both devices i correctly see the numeric keyboard, but with a `,` instead of a `.` for decimals (that's not accepted by the field).
No issues on Android, numeric keyboard has both `,` and `.`. Decimal point is added only if user taps on `.` (not so elegant, but it works all in all)
IMG_6370
Screenshot 2025-03-12 alle 22 37 06

bongio94:
Hmm that's strange. I thought english speaking countries used the `.` as a decimal separator. Is it only the language that it's set to English or the region of the device as well?

But still, this is an issue as it won't work for everyone who sees a `,` instead of the `.`.

A potential solution could be to allow for both symbols (1 time only) but replace the comma with the period when typed in the input field.

Yesterday I saw it's possible to create your own `FilteringTextInputFormatter` so I can try specifying this custom rule.

On a separate note, didn't get what you mean with this:

> Decimal point is added only if user taps on . (not so elegant, but it works all in all)

Isn't this the intended behaviour?

mikev-cw:
> Edit: Oh I think I get it now, given the comma is not accepted the user can only enter the period. This what you meant right?

Correct. Since the keyboard prompted with a `,` AND a `.`, as an italian user I'll have the tendency to tap the comma and nothing happens, that's the point (no pun intended).

I both my iOS devices I have region Italy, that's maybe the issue.

What about "intercept" the input, and put a `.` if the user taps to `,`?

As a side note, we were speaking with @lucaantonelli about the `intl` Flutter package, that with `NumberFormat.currency()` function should address this kind of issues, maybe you want to give it a try?

bongio94:
Yes, it is region related. I just tried switching my Simulator region to Italy (keeping the language US) and it shows the comma.

Yep, I'll make it work with both symbols 👍

lucaantonelli:
@mikev-cw  I think we can go with the currency numberformat later, for now a filtering with regex like that `RegExp(r'^\d+([.,]?\d{0,2})')` accepting both , or . would be fine, as proposed by @bongio94

bongio94:
I've tried the `NumberFormat` from `intl` but it would require a bit more thought as the `amountController` is being edited from outside the `AmountWidget`. 

I'll follow @lucaantonelli suggestion here.

One thing I noticed is that the original regex `RegExp(r'\d*\.?\d{0,2}')` is meant to allow for 2 decimal digits only. Is that how it's supposed to work?

mikev-cw:
> One thing I noticed is that the original regex `RegExp(r'\d*\.?\d{0,2}')` is meant to allow for 2 decimal digits only. Is that how it's supposed to work?

Yes, it is. I cannot imagine a scenario where the user needs more than 2 decimal digits with the 4 currencies we give support to

bongio94:
I'll add that in as well. 
I think it's not working at the moment (you should be able to reproduce this from main as well)

bongio94:
@mikev-cw I've pushed the changes 🙂

mikev-cw:
Thank you, I will review likely tomorrow.
Meanwhile, i don't know if you read the edited text on my prev comment: if you like, push another commit adding your name on the collaborators page :)

mikev-cw:
Working as expected. Very good work!

PR #258, opened 2025-03-14
Fix a Typo, there wasn't a newline where needed

simo498:


napitek:
@theperu this can also be merged

Issue #245, opened 2025-03-13
Problems in the "Transfer" page


federicopozzato:
In the transfer page the field below "from" should have a dynamic icon that switch color and icon based on the account I chose. Same for the "to" field. 
Here the "active state": https://www.figma.com/design/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?node-id=8095-7620&t=wIcBgzLJtYVqsUqS-11

The text should be in an active state when an account is selected (https://www.figma.com/design/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?node-id=1482-4204&t=wIcBgzLJtYVqsUqS-11)

Hope is clear!

![Image](https://github.com/user-attachments/assets/e572c3d2-b9c2-4f08-bcac-9993e7df41d2)

PR #260, opened 2025-03-14
fix: Update transfer page UI

dhruvanbhalara:
- Added Source/Destination Account icon/color while selecting data
- Closes https://github.com/RIP-Comm/sossoldi/issues/245


mikev-cw:
Thank you for this contribution!
For the next time, please, open an issue first, so we can assign it and avoid multiple devs working on the same path.

IDK if is a related issue, but transfer is not working: i have always grey button and can't add it. Is the same for you?

PR #272, opened 2025-03-14
chore: add Docker image to develop inside a Docker container in VSCode (Android)

davfru:
(Android only)

Instead of manually setting up the required dependencies locally, this PR introduces a Docker-based development environment. The provided Docker image includes all necessary dependencies, making it more convenient to develop for Android without modifying the local setup. 

Additionally, this PR includes instructions on how to use the Docker environment within VSCode.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,16 @@
+FROM ghcr.io/cirruslabs/flutter:3.29.1
+
+# Set up Android NDK
+ENV ANDROID_NDK_VERSION=22
+ENV ANDROID_HOME=/opt/android-sdk-linux
+ENV ANDROID_NDK_HOME=$ANDROID_HOME/ndk/$ANDROID_NDK_VERSION
+
+# Download and install the NDK
+RUN mkdir -p $ANDROID_HOME/ndk && \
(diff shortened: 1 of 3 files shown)


napitek:
I must say that the idea can fit, I like it.
The problem is that there is no unique setup for working on the app and there are some flaws:

- **Fixed NDK version**: You're using a specific NDK version (22) which might become outdated. It would be better to parameterize the version with an ARG to make future updates easier.

- **No download verification**: There's no verification of the downloaded file's integrity (like using SHA256 checksums).

- **WORKDIR**:  although it is less important, a `WORKDIR` in Dockerfile would not be bad. You could almost think of using multiple versions for testing in the same container, or one container per version with the dedicated `WORKDIR`
I think your flutter version uses `/build` as default path.

napitek:
@davfru 
Do we want to discuss the use of docker for the development environment?
It is certainly not currently a priority based on the timeline, but in the future it could certainly be a viable alternative to set up an environment that works for everyone, despite the limitations of the case

davfru:
Hi @napitek, thanks for your feedback. I'll work on them soon.

davfru:
Hi @napitek

I made some changes based on your feedback, let me know what you think.

mikev-cw:
> @davfru 
> 
> Do we want to discuss the use of docker for the development environment?

Yes we do. It's on the agenda of our next meeting later this week 👍🏻

PR #276, opened 2025-03-14
feat: Resolve #200 - Added confirmation dialog when deleting an account

### LINKED ISSUE #200: Bug: No confirmation message is shown when deleting an account

bongio94:
In this PR:

- Added a new widget, `AdaptiveDialog`:
This was done for having a convenient way of displaying a native looking dialog. The problem with `AlertDialog.adaptive` is that you still need to differentiate the actions, otherwise you'll end up with Material buttons on the Cupertino widgets. To simplify this process I've create a convenience class for the actions, which will be mapped to the correct platform widget (`TextButton` on Android, `CupertinoDialogAction` on iOS).

- Added the confirmation dialog:
When tapping on the _Delete account_ button you'll now see a dialog which will ask you to confirm your choice. I don't know if the wording on the dialog can be improved, it looked good enough for me but I'm open to suggestions.

- After the account has been deleted, navigate back to the `'/account-list'` route.

- Created the widgets folder in `lib/pages/accounts` and added the `ConfirmAccountDeletionDialog`.

- `TextButton` now uses the _NunitoSans_ font:
Please let me know if using _Roboto_ was a design choice. If not, by specifying the `fontFamily` in the `Theme` there's no need of assigning a `textTheme` to each `Text` in `TextButton` (unless you want to use a different `textTheme` than the default one of course).

- Added some basic tests for the `AdaptiveDialog` widget.

| | Light mode | Dark mode |
|-|------------|-----------|
| Android |Screenshot_1741992130 |Screenshot_1741992120 |
| iOS |Simulator Screenshot - iPhone 16 Pro - 2025-03-14 at 23 41 40 |Simulator Screenshot - iPhone 16 Pro - 2025-03-14 at 23 41 52 |

Resolve #200

### DIFF
--- a/lib/utils/app_theme.dart
+++ b/lib/utils/app_theme.dart
@@ -78,6 +78,7 @@ class AppTheme {
         ),
         textStyle: WidgetStatePropertyAll(
           TextStyle(
+            fontFamily: 'NunitoSans',
             fontSize: 14.0,
             fontWeight: FontWeight.w700,
           ),
@@ -246,6 +247,7 @@ class AppTheme {
         ),
         textStyle: WidgetStatePropertyAll(
           TextStyle(
+            fontFamily: 'NunitoSans',
             fontSize: 14.0,
             fontWeight: FontWeight.w700,
           ),
--- a/lib/utils/app_theme.dart
+++ b/lib/utils/app_theme.dart
@@ -53,7 +53,7 @@ class AppTheme {
         shape: WidgetStatePropertyAll(
           RoundedRectangleBorder(borderRadius: BorderRadius.circular(8)),
         ),
-        textStyle: WidgetStatePropertyAll(
+        textStyle: const WidgetStatePropertyAll(
           TextStyle(
             fontFamily: 'NunitoSans',
             fontSize: 18.0,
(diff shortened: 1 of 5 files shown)


lib/utils/app_theme.dart
```diff
@@ -55,6 +55,7 @@ class AppTheme {
         ),
         textStyle: const WidgetStatePropertyAll(
           TextStyle(
+            fontFamily: 'NunitoSans',
```
lucaantonelli, line 58: is it necessary to explicitly specify the font? I tried without and it seems there is still the right font. Let me know :)

bongio94: I can see that _Roboto_ is used if the `fontFamily` is removed

Screenshot_1742033067

lucaantonelli: Sorry you're right, since this problem applies to elevated button too, can you add the fontfamily also on their textstyle in app theme?

bongio94: Done 🙂

PR #278, opened 2025-03-15
feat: allow user to enable local device authentication

### LINKED ISSUE #248: Implement local authentication with biometric support
One feature that I believe should be implemented is local authentication. This includes authentication with biometrics such as fingerprint or facial recognition. This functionality would significantly improve security and user experience by reducing the need to enter complex passwords.

gabrielecabrini:
Added support for local device authentication, by default this option defaults to false.

This closes #248

Screenshot 2025-03-15 012726
Screenshot 2025-03-15 012814
Screenshot 2025-03-15 012827

### DIFF
--- a/android/app/build.gradle.kts
+++ b/android/app/build.gradle.kts
@@ -53,10 +53,10 @@ android {
 
     signingConfigs {
         create("release") {
-            keyAlias = keystoreProperties["keyAlias"] as String
-            keyPassword = keystoreProperties["keyPassword"] as String
+            keyAlias = keystoreProperties["keyAlias"] as String?
+            keyPassword = keystoreProperties["keyPassword"] as String?
             storeFile = keystoreProperties["storeFile"]?.let { file(it) }
-            storePassword = keystoreProperties["storePassword"] as String
+            storePassword = keystoreProperties["storePassword"] as String?
         }
     }
 
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -11,8 +11,10 @@
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
     <uses-permission android:name="android.permission.USE_FULL_SCREEN_INTENT"/>
 
-		<!-- Storage permissions -->
+	<!-- Storage permissions -->
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" android:maxSdkVersion="32"/>
+    <!-- Handle authentication -->
+    <uses-permission android:name="android.permission.USE_BIOMETRIC"/>
 
     <application
         android:label="Sossoldi"
--- a/android/app/src/main/kotlin/com/ripster/sossoldi/MainActivity.kt
+++ b/android/app/src/main/kotlin/com/ripster/sossoldi/MainActivity.kt
@@ -1,5 +1,5 @@
 package com.ripster.sossoldi
 
-import io.flutter.embedding.android.FlutterActivity
+import io.flutter.embedding.android.FlutterFragmentActivity
 
-class MainActivity : FlutterActivity()
+class MainActivity: FlutterFragmentActivity()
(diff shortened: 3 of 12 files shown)


bongio94:
Hello @gabrielecabrini 

This looks good to me, only a couple of things that might be improved imo (please consider that I've never used the local_auth package and my assumptions might be wrong!)

- I've noticed that `if (!didAuthenticate) return;` it just stops the execution by exiting from `main()` and the app remains on the splash screen. Wouldn't it be better if the app was closed altogether when the user fails to authenticate? In my test I've just tried to tap outside of the authentication bottom sheet. See attached. I did some googling and it seems like the preferred method for terminating the app from the code is `SystemChannels.platform.invokeMethod('SystemNavigator.pop');`, docs [here](https://api.flutter.dev/flutter/services/SystemNavigator/pop.html).  
Also, it seems like Apple is does not allow for the developer to terminate the app programmatically from the code (you could do it by calling `exit(0)` but Apple might take the app down from the AppStore for this reason, so I think it's better if we don't try that 😅 ). Another option that it might be used on iOS is to move the app in background if the authentication fails. But I don't know how is the app behaving when the authentication fails on iOS, I have only tested on Android.

- The second thing is: I've noticed the `pubspec.lock` contains lots of new additions. Besides the obvious `local_auth` dependencies which have to be there, are the others all transitive dependencies from `local_auth`? I've had a look at the dependencies of `local_auth` from [its pub.dev page](https://pub.dev/packages/local_auth) (you can find this section at the end of the right info panel) and I can't match them with the ones that are being added here. Could you please double check that all of the other dependencies added to the `pubspec.lock` are required?


https://github.com/user-attachments/assets/9a39500d-b77d-41a2-801e-cef9218b8856

gabrielecabrini:
Hi, I don't think that's a valid thing to close / background the app if the authentication goes wrong, we could add a route saying "Authentication required" with a button that will retry authentication, like WhatsApp does for example.

For the `pubspeck.lock` file, i think that's just its dependencies

bongio94:
I think that a page that let the user retry the authentication would definitely be better 👍 

@federicopozzato is it worth to discuss this as well in today's UI/UX meeting?

gabrielecabrini:
image

For instance, this is how the page I want to implement looks on WhatsApp

bongio94:
Thanks for providing this example.
I think it would be best to wait for inputs from the design team on this 👍

bongio94:
I don't know if I'll make it for this evening meeting 🙁

gabrielecabrini:
Sorry I couldn't make it to the meeting, I'm waiting for info's about that page :)

kommpn:
hi, this feature is very interesting. is there any update? do you need some help to implement this feature @gabrielecabrini ?
thanks :)

gabrielecabrini:
Hi, there's already a PR for this, we are waiting the design team 😄

gabrielecabrini:
I think we're waiting the design team for this

federicopozzato:
Yes, sorry guys, we are a bit late with this, and I don't think will be ready soon

gabrielecabrini:
sure no problem

theperu (changes requested):
Hi @gabrielecabrini! Here is what I found:

## Android
I noticed an issue if the phone as no pin/sequence to unlock it, in this situation the app does not open and I can't access it.
We should probably have a way to check if no pin is needed to unlock the phone and in this case we should just skip the check.

## iOS
It seems fine also but I only have the simulator at the moment and not a real device, it would be ideal if you could record how it works on iOS if you activate it when you have a pin and when you don't have it

gabrielecabrini:
I've fixed the issue. the `isDeviceSupported()` method returns false if the device hasn't any sort of authentication in place, be it biometrics or PIN.
I can't test it right now on iOS, but this will also allow iOS to bypass auth if not present

theperu:
Understood, I'll look for someone with an iOS device that could review this as well 😉

PR #289, opened 2025-03-15
Added category type icon ( import/export ) on category list

lucafulgenzi:
Added the category type icon ( income/expense) to the end of the category row.
In addition I modified the RoundedIcon component by adding the optional color property that as a fallback uses white.
image
image


theperu:
Hi @lucafulgenzi could you open an issue about this? I would like to discuss the implementation with the designing team first and then decide which is the best approach to convey this info

PR #290, opened 2025-03-15
fix: Resolve #251 - Ask user to create a category before adding a category budget

### LINKED ISSUE #251: Android: Can't create a new Budget Category or Category
On a empty istance of the app (or with a created account) it is not possible to create a new category budget from Planning > + New budget (see attached file) or from the New Transaction menu (the + button).

Phone details:
- One UI Version: 6.1
- Android Version: 14
- Google Play System Update: June 1, 2024
- Baseband Version: S926BXXS5AYB2
- Kernel Version: 6.1.75-android14-11
- Build Version: UP1A.231005.007.S926BXXS5AYB2

https://github.com/user-attachments/assets/35a75921-7e67-4b47-b285-8f9de87b52ea

bongio94:
As flagged in #251, when there are no categories and the user tries to add a category budget, an exception is thrown (and the user is unable to create a category budget).

To solve this issue I've added a check on `categories`, if empty, shows a `SnackBar` which asks the user to add a category before adding a category budget:
  - The `SnackBar` has an _Add_ button. Tapping on the button shows the `/add-category` route. 
  - If the user adds a category, the `/add-category` route is popped and `true` is returned as a `result` from the `.pop()`. If the user doesn't add a category, `false` or `null` is returned as a `result` (depending on how the user dismisses the  `/add-category`).
  - When a category is added successfully, `_loadCategories()` is called.
 
Other changes:
  - Added arguments handling on the `/add-category` route definition.
  - Wrapped `lib\pages\planning_page\manage_budget_page.dart` with `Scaffold`enabling access to `ScaffoldMessenger.of(context)` for proper `SnackBar` handling.
  - Added `MissingCategorySnackBarContent` to separate the visual side of the `SnackBack`, not really relevant, just a bit cleaner I guess. It's probably worth to standardize `SnackBars` into a centralized, configurable `Widget`.

## Demo:
https://github.com/user-attachments/assets/1f88c812-49a8-4124-80f4-32965e92e2bd

Resolve #251


napitek:
@bongio94 great work! Thanks.
This solves some problems when the user of the app wants to start from 0 and does not yet have defined categories.
There will be to fix the color of the `SnackBar` for the light theme (is currently dark blue because of `colorScheme.onPrimary`), but @mikev-cw , for me, we can merge

bongio94:
Thanks for the review @napitek 🙏 

I think the color is correct (the dark background on light mode and the light background on dark mode), the [M3 Specs for SnackBar](https://m3.material.io/components/snackbar/specs#3c1ed6ed-2314-408b-aa75-634de186ec0a) shows that `inverseSurface` is used for the background color and this is applied to Flutter as you can see here:
https://github.com/esouthren/flutter/blob/dbcff31ed0a388b905009bf22c8226ed01a4bff5/packages/flutter/lib/src/material/snack_bar.dart#L900

My assumption here is that `SnackBar`s could be missed if there wasn't enough contrast with the rest of the UI, that's why they're giving them the `inverseSurface` color.

I'm wondering now, which color is actually applying? 🤔 
`inverseSurface` is not defined in the `customColorScheme` and that dark blue seems like a color in the sossoldi palette!

I'm happy to discuss a centralized theming for `SnackBar`s 🙂

napitek:
yes it is not a problem that directly affects the issue. I would say we can talk about it when the `SnackBar` part is discussed with the designers.

napitek:
i think we can merge this? @bongio94

bongio94:
Is this okay from the design side?
@federicopozzato

federicopozzato:
Yes!

bongio94:
I can see some issues with the `SnackBar`.

Comparing the original version shown [here](https://github.com/RIP-Comm/sossoldi/pull/290#issue-2922075302), it has now changed to this:


https://github.com/user-attachments/assets/3ac9c910-38d8-4934-8c2f-815ed52ce610

Because of some changes that were made in `lib\utils\app_theme.dart`.
This is out of the box, without updating the `showSnackBar` function.

If I use this function instead (`lib\utils\snack_bars\snack_bar.dart`) I got this (watch till the end for seeing the Flutter inspector):

https://github.com/user-attachments/assets/f8fc7f87-1126-4822-a987-b0c59ce97d1d

I feel like these updates to the `SnackBar` theme make snack bars difficult to see as there is no contrast with the background.

As mentioned in [this](https://github.com/RIP-Comm/sossoldi/pull/290#issuecomment-2741858427) post, Material adopts the same styling strategy.

My proposal is to update the `SnackBar` theme with these values:

- `backgroundColor`: from `primaryContainer` to `inverseSurface`
- `color` (of `TextStyle`): from `onSurface` to `onInverseSurface`

This updates the `SnackBar` styling to:

https://github.com/user-attachments/assets/74b4dc83-4a2d-40da-a3eb-2bc0631c106c


Also, I would remove the padding [here](https://github.com/bongio94/sossoldi/blob/30b51712a2a72c64205dc1ba0b948fa035cb6027/lib/utils/snack_bars/snack_bar.dart#L36) and also the overflow [here](https://github.com/bongio94/sossoldi/blob/30b51712a2a72c64205dc1ba0b948fa035cb6027/lib/utils/snack_bars/snack_bar.dart#L19) as this is cutting the text to one line.

@federicopozzato Sorry to drag you into this again but we'd really need some guidance from the design side. I don't think I've seen` SnackBar`s as UI elements in Figma

cc @fres-sudo

federicopozzato:
No, that component is not present in figma. It’s what done! Obviously the version more contrasted is better. 
Go on with that. I will add it on figma for the DS in the following iterations

theperu:
@bongio94 I don't know if this is needed but if you want we could use a shorter message which might fit better in the SnackBar, in this case I would propose something like `Add a category first to set a budget` or `Can't add budget without a category`

bongio94:
@theperu I wouldn't want to limit the `SnackBar`'s messages to one line to be honest. 

A surprisingly large amount of people use enlarged text size on their devices and even a short String might be cut with ellipsis in these instances.

fres-sudo:
> @theperu I wouldn't want to limit the `SnackBar`'s messages to one line to be honest.
> 
Even if I was the author of the `SnackBar`, I do agree to increase the lines limit of the SnackBar to at least 2 lines.

bongio94:
I can keep the `overflow` but set `maxLines` to 2 @fres-sudo 

Btw, format job is failing 🤔 
`dart format .`  is returning `Formatted 126 files (0 changed) in 1.09 seconds.` so I guess there's nothing to format

bongio94:
I'll push [these](https://github.com/RIP-Comm/sossoldi/pull/290#issuecomment-2779400147) changes, please have a look at how these look on you emulator/simulator

bongio94:
@fres-sudo @theperu This is ready to be reviewed 🙂

lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -44,6 +45,33 @@ class _ManageBudgetPageState extends ConsumerState<ManageBudgetPage> {
...
+          context,
+          '/add-category',
+          arguments: {'hideIncome': true},
+        ) as bool?;
+
+        // Only call _loadCategories() if a new category has been added.
+        //
+        // When the user navigates back without adding a category without pressing the back button
+        // on the /add-category page, Navigator will return null, which will be handled as a false.
+        if (categoryAdded ?? false) _loadCategories();
```
fres-sudo, line 70: I think all this context is unnecessary. I can easily tell what is going on just by looking at the code. Also, the declaration before the assignment is redundant. Why can't we just do this:

```dart
final categoryAdded = await Navigator.pushNamed(
          context,
          '/add-category',
          arguments: {'hideIncome': true},
        ) as bool?;
if (categoryAdded ?? false) _loadCategories();
```
By reading at this code I can tell that the variable is a `bool` or `null`, I can tell that we're pushing the `/add-category` route and awaiting its result. The variable name is telling me what we're expecting, and I can tell that if a category has been added, then I need to reload them.

lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -52,127 +80,137 @@ class _ManageBudgetPageState extends ConsumerState<ManageBudgetPage> {
...
-                        active: true,
-                        amountLimit: 100,
-                        idCategory: categories[0].id!,
-                        name: categories[0].name));
+                    widget.onRefreshBudgets();
+                    Navigator.of(context).pop();
                   });
                 },
-                label: Text("Add category budget"),
+                style: ElevatedButton.styleFrom(elevation: 2),
```
fres-sudo, line 207: Genuinely wondering, why this `elevation`?

lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -52,127 +80,137 @@ class _ManageBudgetPageState extends ConsumerState<ManageBudgetPage> {
...
+                    await BudgetMethods().insertOrUpdate(item);
+                  }
                   setState(() {
-                    budgets.add(Budget(
-                        active: true,
-                        amountLimit: 100,
-                        idCategory: categories[0].id!,
-                        name: categories[0].name));
+                    widget.onRefreshBudgets();
+                    Navigator.of(context).pop();
```
fres-sudo, line 204: Is that safe? Can't we just move out the `setState`, nothing will ever happen...but you'll never know

lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -52,127 +80,137 @@ class _ManageBudgetPageState extends ConsumerState<ManageBudgetPage> {
...
+          Text(
+            "Your monthly budget will be: ${budgets.isEmpty ? 0 : budgets.fold(0, (sum, e) => sum + e.amountLimit.toInt())}€",
+            style: Theme.of(context).textTheme.titleMedium,
+          ),
+          const SizedBox(height: 16),
+          const Divider(indent: 16, endIndent: 16),
+          Padding(
+            padding: const EdgeInsets.all(16.0),
+            child: SizedBox(
+              width: double.infinity,
```
fres-sudo, line 193: Can you avoid this and just wrap the `child` with a `Center`?

lib/pages/planning_page/widget/missing_category_snackabr.dart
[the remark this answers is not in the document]
bongio94: This is a leftover from the original implementation, before the `showSnackBar` was implemented, I'll remove it 👍

lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -44,6 +45,33 @@ class _ManageBudgetPageState extends ConsumerState<ManageBudgetPage> {
     await ref.read(budgetsProvider.notifier).refreshBudgets();
   }
 
+  void handleEmptyCategories() {
+    showSnackBar(
+      context,
+      message: "Add a category first to set a budget",
+      actionLabel: "ADD",
+      onAction: () async {
+        ScaffoldMessenger.of(context).clearSnackBars();
```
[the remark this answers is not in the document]
bongio94: Same as [this](https://github.com/RIP-Comm/sossoldi/pull/290#discussion_r2040568252) one.

lib/pages/planning_page/manage_budget_page.dart
bongio94: I can move the declaration in line with the assignment 👍 

On the comments, I get this is a simple piece of code but I'd rather have commented code than no comments at all!

This is an open source project and not everyone is at the same skill level, it might be obvious for you, but not for everyone 🙂

lib/pages/planning_page/manage_budget_page.dart
bongio94: Not part of my changes, already present in main:

https://github.com/RIP-Comm/sossoldi/blob/993a7c375b458bcb749f6f7473ac693ef1469885/lib/pages/planning_page/manage_budget_page.dart#L170

lib/pages/planning_page/manage_budget_page.dart
bongio94: Same as the above:

https://github.com/RIP-Comm/sossoldi/blob/993a7c375b458bcb749f6f7473ac693ef1469885/lib/pages/planning_page/manage_budget_page.dart#L165-L168

lib/pages/planning_page/manage_budget_page.dart
bongio94: Same as the above:

https://github.com/RIP-Comm/sossoldi/blob/993a7c375b458bcb749f6f7473ac693ef1469885/lib/pages/planning_page/manage_budget_page.dart#L155-L156

lib/pages/planning_page/manage_budget_page.dart
fres-sudo: Yes, I knew that, but those kinds of issues I think that are appropriate to fix in that place, even if they're not part of your diff or scope of the PRs, otherwise someone else has to do that job in another place.

bongio94: Tbh, I disagree with this.

Changes in a PR should adhere to the scope of the PR.

The risk is to add clutter and make it less readable and harder to review. 
Additionally, there's the chance of loosing track of the changes.

Also, I wouldn't know what to change this to without an input from the design team.

fres-sudo (approved):

bongio94:
@fres-sudo 

I've rebased with main 👍 

I've also implemented your suggestions:

- got rid of the comments
- removed the `setState`
- removed `double.infinity` and used `CrossAxisAlignment.stretch` to make the button full width
- removed the `elevation` from the button

Something I added is a `SafeArea` as the button was very close to the navigation bar.

If you got time please have a look at these changes and see if I've missed something 🙂 

Thank you!

lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -53,125 +70,139 @@ class _ManageBudgetPageState extends ConsumerState<ManageBudgetPage> {
...
+              child: Text(
+                "Select the categories to create your budget",
+                style: Theme.of(context).textTheme.titleLarge,
               ),
-              Text(
-                'AMOUNT',
-                textAlign: TextAlign.right,
+            ),
+            const Padding(
+              padding: EdgeInsets.symmetric(horizontal: 16, vertical: 8),
```
fres-sudo, line 86: Same here: 

```dart
const Padding(
              padding: EdgeInsets.symmetric(horizontal: Sizes.lg, vertical: Sizes.sm),
```

lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -53,125 +70,139 @@ class _ManageBudgetPageState extends ConsumerState<ManageBudgetPage> {
...
+              child: Column(
+                children: [
+                  ListView.builder(
+                    shrinkWrap: true,
+                    itemCount: budgets.length,
+                    itemBuilder: (context, index) {
+                      return Dismissible(
+                        key: Key(budgets[index].idCategory.toString()),
+                        background: Container(
+                          padding: const EdgeInsets.only(right: 20.0),
```
fres-sudo, line 111: here we can use:

```dart
EdgeInsets.only(right: Sizes.lg),
```

lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -53,125 +70,139 @@ class _ManageBudgetPageState extends ConsumerState<ManageBudgetPage> {
...
-                        updateBudget(updatedBudget, index);
-                      },
-                    ),
-                  );
-                },
+                  ),
+                  SizedBox(height: 8),
+                  Text("Swipe left to delete",
+                      style: Theme.of(context).textTheme.bodySmall),
+                  SizedBox(height: 12),
```
fres-sudo, line 151: here:

```dart
SizedBox(height: Sizes.md),
```

lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -53,125 +70,139 @@ class _ManageBudgetPageState extends ConsumerState<ManageBudgetPage> {
...
+                          );
+                        },
+                      );
+                    },
+                    label: Text("Add category budget"),
+                  ),
+                ],
+              ),
+            ),
+            const SizedBox(height: 16),
```
fres-sudo, line 175: here:
```dart
SizedBox(height: Sizes.lg),
```

lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -53,125 +70,139 @@ class _ManageBudgetPageState extends ConsumerState<ManageBudgetPage> {
...
-                onPressed: () {
-                  setState(() {
-                    budgets.add(Budget(
-                        active: true,
-                        amountLimit: 100,
-                        idCategory: categories[0].id!,
-                        name: categories[0].name));
-                  });
+            ),
+            const SizedBox(height: 16),
```
fres-sudo, line 182: here:
```dart
SizedBox(height: Sizes.lg),
```

lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -53,125 +70,139 @@ class _ManageBudgetPageState extends ConsumerState<ManageBudgetPage> {
...
+            const Divider(indent: 16, endIndent: 16),
+            Padding(
+              padding: const EdgeInsets.all(16.0),
```
fres-sudo, line 185: here:
```dart
padding: const EdgeInsets.all(Sizes.lg),
```

fres-sudo (changes requested):
Thanks for the rebase!
I see that during the conflict resolution you had to "sacrifice" some part of the old code that had the correct Sizing properties, so I came to your rescue with that reviews.

bongio94:
@fres-sudo Everything should be using the `Sizes` class now, thanks for the review 👍

fres-sudo (approved):
Thanks!

napitek:
@theperu this can be merged

Issue #244, opened 2025-03-13
"Add transaction" button behaviour when typing


federicopozzato:
When I'm typing the description, the "Add transaction" button should not be above the keyboard

![Image](https://github.com/user-attachments/assets/94d6e77a-1048-4267-9fff-7b437b01f9f7)

theperu:
This might be closely related to #242 let's keep an eye on it

gohaneri:
Opened a pull request to fix that

Issue #242, opened 2025-03-13
Transaction page behaviour


federicopozzato:
The transaction page should be a full-screen modal. (Now it's a page sliding from right)
Reference: https://www.figma.com/design/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?node-id=2982-6443&t=bN6XCLSFk08t2Knh-11

gohaneri:
Opened a pull request to fix that

PR #295, opened 2025-03-15
Fix for #242 and #244

gohaneri:
Fix for #242
Fix for #244


theperu:
Thanks for the contribution! We'll try to review this in the next days. Could you add a comment in each issue that was fixed so that I can assign both to you?

Also for the next contributions it would be best if you commented on the issue before starting to work on it, in this way we can avoid having multiple people working on the same task 😁

PR #298, opened 2025-03-15
Fix analysis warning

federicoviceconti:
This commit solves:
- avoid_print
- BuildContext across async gaps
- 'rethrow' to rethrow a caught exception

### DIFF
--- a/lib/utils/csv_file_picker.dart
+++ b/lib/utils/csv_file_picker.dart
@@ -27,7 +27,7 @@ class CSVFilePicker {
   static Future<File?> pickCSVFile(BuildContext context) async {
     bool permissionGranted = await _requestStoragePermission();
     if (!permissionGranted) {
-      if(context.mounted) {
+      if (context.mounted) {
         showSnackBar(
           context,
           message: 'Storage permission is required',
@@ -47,7 +47,7 @@ class CSVFilePicker {
         return File(result.files.first.path!);
       }
     } catch (e) {
-      if(context.mounted) {
+      if (context.mounted) {
         showSnackBar(
           context,
           message: 'Error picking file: ${e.toString()}',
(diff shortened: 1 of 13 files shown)


lib/utils/csv_file_picker.dart
```diff
@@ -71,19 +75,23 @@ class CSVFilePicker {
...
-        ),
-      );
+      if(context.mounted) {
+        ScaffoldMessenger.of(context).showSnackBar(
+          SnackBar(
+            content: Text('Error saving file: ${e.toString()}'),
+            backgroundColor: Colors.red,
+          ),
+        );
+      }
```
fres-sudo, line 94: In #234 a general way to handle snack bars has been added, can you please refine this code to suit the standard of the codebase?

In this case `showSnackBar()` must be used to ensure proper styling

fres-sudo (changes requested):
I like the `dev.log`and the `rethrow` I leave just a small review regarding the snack bars.

Ik that it wasn't part of your PR, but since you touched also that code I think it's appropriate to refactor it.

lib/utils/csv_file_picker.dart
federicoviceconti: Done. Check it if it works.

fres-sudo: Seems good to me, the only thing that do not convince me is the non-sanitizing of the error. I mean that, rn we're giving bold errors to user, without knowing what we're actually showing up.

But this is a thing for the next big refactoring, in which we'll implement a proper error handling.

For now, it's LGTM 👍

lucaantonelli:
Hi, i think the code is good, we have some conflicts with the latest pr we merged, should not be a big work to solve it with a rebase. Thanks in advance! :)

federicoviceconti:
> Hi, i think the code is good, we have some conflicts with the latest pr we merged, should not be a big work to solve it with a rebase. Thanks in advance! :)

I aligned my code to the latest commit on main. Check if this works for you :)

lucaantonelli:
> > Hi, i think the code is good, we have some conflicts with the latest pr we merged, should not be a big work to solve it with a rebase. Thanks in advance! :)
> 
> I aligned my code to the latest commit on main. Check if this works for you :) @lucaantonelli

It seems you need to run a `dart format .` that's the only thing, i will merge after that :)

Issue #253, opened 2025-03-13
Implementation of FVM (Flutter Version Manager)


VoidElle:
I think [FVM](https://fvm.app/) would be a valuable addition to the project.

It ensures that the Flutter version used is fixed and project-specific rather than system-dependent, 
reducing potential compatibility issues between contributors.

If you guys want, i can work on the implementation, 
just let me know the specific desired flutter version.

uveshm003:
I think 3.24 will be good as I believe that's the last stable version but if we are only targeting application than 3.29.2 is also good.

gabrielecabrini:
I'm using latest 3.29.2 without any problems

marcoredz:
I suggest to use Puro instead: https://puro.dev/

VoidElle:
I actually have no experience with Puro (I wasn’t even aware of its existence, to be honest). 
Based on the download count on its pub.dev page, it appears to be much less popular.

I believe FVM would be a better fit for this project due to its popularity and stability.
I am also not having any issues using the project with the 3.29.2 version.

@theperu If you’d like to assign me the issue, I can implement FVM with Flutter version 3.29.2.

theperu:
@VoidElle I can assign this to you but let's discuss this in next week's call before merging (https://discord.gg/JvSgua28?event=1350169879612887040)

VoidElle:
Got it! :)

marcoredz:
@VoidElle just for knowledge: [Dart 3: Using Puro to manage Flutter versions](https://www.youtube.com/watch?v=olqmK-opexA)

PR #319, opened 2025-03-16
Implementation of FVM (Flutter Version Manager)

VoidElle:
As per #253


fres-sudo:
Seems LGTM to me 👍

dariowskii:
@VoidElle @fres-sudo 

I believe that from the moment you implement FVM, it is also okay to recreate the native project folders with the Flutter cli `3.29.2` version, then doing a merge of any differences (various implementations).

This way you would have the projects with the updated configurations

VoidElle:
Hmm… I’m not sure about the exact version the project was originally built with. If it’s around 3.29, I don’t think it’s necessary. On the other hand, if it’s lower, I think doing that would be a great idea. @theperu, do you have this info?
(Feel free to disagree, I'm not aware of major changes on latest flutter releases)

fres-sudo:
>  I don’t think it’s necessary.

I actually disagree on that, I think @dariowskii is right on that one, and I compleetely missed it.

In the newer version of flutter, the `android` folder has the new `.kts` file (that our `android/` doesn't have) i.e.:

- `build.gradle` has become `build.gradle.kts`

VoidElle:
> (Feel free to disagree, I'm not aware of major changes on latest flutter releases)

My disclaimer was necessary 😅
Okay, I'll work on it!

VoidElle:
@fres-sudo @dariowskii i recreated the native folders using:  
`fvm flutter create . --platforms=android,ios`.

iOS was already aligned.

Android was only misaligned with the .gradle files. 
They have been ported to .gradle.kt files in the latest commits, keeping also the same dependencies versions and functionalities (desugaring / proguard).

dariowskii:
@VoidElle LGTM 🚀

dariowskii:
@VoidElle @fres-sudo I was noticing that the only thing left to do, would be to update the github action to use fvm instead of hardcoding the version.

You could update the `flutter-ci.yml` and `test.yml` file like this:

```yaml
    ...
     - name: Setup FVM
        uses: kuhnroyal/flutter-fvm-config-action/config@v3
        id: fvm-config-action
      
      - name: Setup Flutter
        uses: subosito/flutter-action@v2
        with:
          flutter-version: ${{ steps.fvm-config-action.outputs.FLUTTER_VERSION }}
```

This way the versions would always be in sync 👍🏼

VoidElle:
I missed that, thanks!
I'll update it tonight :smile:

VoidElle:
Done! 😃

dariowskii:
@fres-sudo I saw that the `android/app/build.gradle.kts` (but also the old one) has this kind of signing configs:

```
 buildTypes {

        release {
            signingConfig = signingConfigs.getByName("debug") 
            isMinifyEnabled = true
            proguardFiles(getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro")
        }

        debug {
            signingConfig = signingConfigs.getByName("debug")
        }

    }
```

Is it intentional to have debug configuration for release as well? And why it was commented on the old one?

fres-sudo:
> Is it intentional to have debug configuration for release as well? And why it was commented on the old one?

I have no Idea, I am not that new to the project, the blame doesn't lie @lucaantonelli prolly knows the answer

lucaantonelli:
>  > Is it intentional to have debug configuration for release as well? And why it was commented on the old one?
> 
> I have no Idea, I am not that old to the project, the git blame doesn't lie so @lucaantonelli prolly knows the answer

I first uncommented it and had problems with the gh actions, maybe could be fixed, the error could be caused by the missing key.properties but should not be committed so i don't know if this could be fixed.

VoidElle:
Mhh...I noticed that also during the porting of the .gradle files, but I didn't reported it because i thought it was off topic to this PR. Is not a better approach to create a specific issue for this after merging this PR?

dariowskii:
> Mhh...I noticed that also during the porting of the .gradle files, but I didn't reported it because i thought it was off topic to this PR. Is not a better approach to create a specific issue for this after merging this PR?

Totally agree! Better to open another issue for this

lucaantonelli:
I will test again today and if all it's fine i will merge it.

PR #325, opened 2025-03-17
Revert "feat: Resolve #200 - Added confirmation dialog when deleting an account"

### LINKED ISSUE #200: Bug: No confirmation message is shown when deleting an account

bongio94:
Reverts RIP-Comm/sossoldi#276

@theperu 
Please accept this revert, wrong account used for some commits.
I'll reopen with commits form the correct account.


lucaantonelli:
Hi Bongio,

I understand your concern and saw the commits. I really appreciate your work and effort, don’t get me wrong. However, accepting reverts and then reapplying the same code just to change the commit author would unnecessarily complicate the repository history without providing a real benefit for us. Additionally, the commits with the wrong account would still be visible in the history, so it wouldn’t fully resolve the issue.

I hope you understand our perspective. Let me know if you’d like to discuss alternative solutions.

PR #301, opened 2025-03-15
chore: Resolve #243 - Improve platform specific UI elements and route transitions

bongio94:
In this PR:

- Replaced `CupertinoSwitch` with `Switch.adaptive` in `lib/pages/accounts/add_account.dart`.
- Added `SwitchThemeData`:
  - @federicopozzato probably worth to have a look at the colors, feel free to suggest changes. 
  You can find more info on the [Material3 Switch guidance](https://m3.material.io/components/switch/overview).
- Created the `buildAdaptiveRoute` wrapper function for `_materialPageRoute` and `_cupertinoPageRoute` in `lib/routes.dart`. This ensure that a Material route is used on Android and a Cupertino route is used on iOS. 
It might give conflicts with #295 and #290.

|LightMode|DarkMode|
|-----------|-----------|
|Screenshot_1742060920|Screenshot_1742060907|

Resolve #243

PR #328, opened 2025-03-18
chore: Resolve #243 - Improve platform specific UI elements and route transitions

### LINKED ISSUE #243: CupertinoSwitch are used in the Edit account page
### Description
While working on #200 I noticed that in the _Edit account_ page (Settings -> Accounts -> tap on one of your accounts) `CupertinoSwitch` are used for the two `ListTile` just above the _Delete account_ button.

### Proposed fix
Replace `CupertinoSwitch` with `Switch.adaptive`.

bongio94:
Copy of #301 

In this PR:

- Replaced `CupertinoSwitch` with `Switch.adaptive` in `lib/pages/accounts/add_account.dart`.
- Added `SwitchThemeData`:
  - @federicopozzato probably worth to have a look at the colors, feel free to suggest changes. 
  You can find more info on the [Material3 Switch guidance](https://m3.material.io/components/switch/overview).
- Created the `buildAdaptiveRoute` wrapper function for `_materialPageRoute` and `_cupertinoPageRoute` in `lib/routes.dart`. This ensure that a Material route is used on Android and a Cupertino route is used on iOS. 
It might give conflicts with #295 and #290.

|LightMode|DarkMode|
|-----------|-----------|
|Screenshot_1742060920|Screenshot_1742060907|

Resolve #243

### DIFF
--- a/lib/pages/accounts/add_account.dart
+++ b/lib/pages/accounts/add_account.dart
@@ -298,9 +298,7 @@ class _AddAccountState extends ConsumerState<AddAccount> with Functions {
                               RegExp(r'^\d*\.?\d{0,2}'),
                             ),
                           ],
-                          style: Theme.of(context)
-                              .textTheme
-                              .titleLarge,
+                          style: Theme.of(context).textTheme.titleLarge,
                         ),
                       ],
                     ),
@@ -327,7 +325,7 @@ class _AddAccountState extends ConsumerState<AddAccount> with Functions {
                                 "Set as main account",
                                 style: Theme.of(context).textTheme.bodyLarge,
                               ),
-                              CupertinoSwitch(
+                              Switch.adaptive(
                                 value: mainAccount,
                                 onChanged: (value) =>
                                     setState(() => mainAccount = value),
@@ -345,7 +343,7 @@ class _AddAccountState extends ConsumerState<AddAccount> with Functions {
                                 "Counts for the net worth",
                                 style: Theme.of(context).textTheme.bodyLarge,
                               ),
-                              CupertinoSwitch(
+                              Switch.adaptive(
                                 value: countNetWorth,
                                 onChanged: (value) =>
                                     setState(() => countNetWorth = value),
--- a/lib/routes.dart
+++ b/lib/routes.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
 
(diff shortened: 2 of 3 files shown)


federicopozzato:
Thanks! Colors look good to me.

fres-sudo:
Seems LGTM to me! 

One small thing, I see that the "issue" is fixed on Android, but on iOS we still see the green background, that okay sticks with apple guidelines but doens't stick with our design system.

So I tought about using our primary color as active color also for the Cupertino Switch if @federicopozzato agrees.

For you @bongio94 the only solution that I found for that is to use an `adaptation` in the `ThemeData` as follows:

```dart
    adaptations: [ SwitchThemeAdaptation() ],
```

the `SwitchThemeAdaptation` class implemetation:

```dart

class SwitchThemeAdaptation extends Adaptation<SwitchThemeData> {
  const SwitchThemeAdaptation();

  @override
  SwitchThemeData adapt(ThemeData theme, SwitchThemeData defaultValue) {
    switch (theme.platform) {
      case TargetPlatform.android:
      case TargetPlatform.fuchsia:
      case TargetPlatform.linux:
      case TargetPlatform.windows:
        return defaultValue;
      case TargetPlatform.iOS:
      case TargetPlatform.macOS:
        return  SwitchThemeData(
          trackColor: WidgetStateProperty.fromMap({
            WidgetState.selected: theme.colorScheme.secondary,
          })
        );
    }
  }
}
```

I already tested it on iOS and it works fine:

<table>
 <tr> 
   <td> Active </td>
  <td>Inactive  </td>
</tr>
  <tr>
    <td valign="top">  </td>
    <td valign="top"></td>
  </tr>
</table>



Wdyt about it?

cc @theperu

bongio94:
Thanks for the feedback @fres-sudo 

I wasn't really planning to customise the `CupertinoSwitch` as well but I agree that giving it the brand identity makes it look better in the app.

If @federicopozzato agrees, I'll add this 👍

federicopozzato:
Thanks for the attention to the detail!
I think is good in both ways: here the goal is to show the user that he activated a switch and now it's active. The user is used to see the switch with the green bg, so it's not really a problem if the color stands out from our color palette.
But now let's keep it the way you did it with the primary color as active. It will be a matter of discussion in the design reviews, and for the future we will provide better guidance for the use of the colors.

Thanks again, for me you can close this issue.

bongio94:
`CupertinoSwitch` `Theme` adaptation has been added. Thanks @fres-sudo for the suggestion!

@theperu @mikev-cw @lucaantonelli feel free to merge 🙂

fres-sudo:
just a small detail...I would personally prefer to move the Adaptation to another file, the theme file seems just enough big. Wdyt?

bongio94:
My reasoning for having it there was to have everything `Theme` related in the same file. 
I thought about creating a separate file with the `Adaptation` but I felt like it wouldn't belong anywhere tbh 😅

Issue #337, opened 2025-03-18
Cannot Create a Budget: Keyboard Covers Save Button


RiSi42:
Description:
When creating a budget, opening the keyboard to enter an amount prevents accessing the Save button, as the keyboard does not close.

Steps to reproduce:
	1.	Open the Budget section
	2.	Create a new budget
	3.	Enter an amount (keyboard opens)
	4.	Notice that the keyboard covers the Save button and cannot be dismissed

Expected result:
There should be a way to close the keyboard, or the Save button should always be accessible.

Actual result:
The Save button is hidden by the keyboard, making it impossible to save the budget.

Environment:
	•	iPhone iOS 18.3.1 e 2

real-Hafez:
I see the issue. Could someone assign it to me?

PR #339, opened 2025-03-19
Fix issue #337 

real-Hafez:
Fix the issue where the keyboard does not appear when adding or editing a budget 
Fix issue #337

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -121,6 +121,14 @@ packages:
       url: "https://pub.dev"
     source: hosted
     version: "1.11.1"
+  cross_file:
+    dependency: transitive
+    description:
+      name: cross_file
+      sha256: "7caf6a750a0c04effbb52a676dce9a4a592e10ad35c34d6d2d0e4811160d5670"
+      url: "https://pub.dev"
+    source: hosted
+    version: "0.3.4+2"
   crypto:
     dependency: transitive
     description:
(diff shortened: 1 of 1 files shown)


lib/pages/planning_page/manage_budget_page.dart
```diff
@@ -51,123 +52,135 @@ class _ManageBudgetPageState extends ConsumerState<ManageBudgetPage> {
...
-        Padding(
-          padding: const EdgeInsets.all(15),
-          child: Text(
-            "Select the categories to create your budget",
-            style: Theme.of(context).textTheme.titleLarge,
+    return AnimatedPadding(
+      padding: EdgeInsets.only(
+          bottom: MediaQuery.of(context)
+              .viewInsets
+              .bottom), // Adjust padding dynamically to show save budget button when keyboord apper
```
fres-sudo, line 59: This comment do not add any value to the code, you can remove it.

fres-sudo (commented):
Didn't test it yet but here some concerns about the code in general:

- Avoid committing generated files like `.lock` and `macos` related files
- A better PR name and description would be appreciated
- What is this "test" commit is for? Notice that, if not being squashed, all the commits will be merged in the main branch of the repo, so please add only meaningful commits. In addition, this "test" commit seems to contain the generated files that I was talking about earlier, it can be removed apparently.

real-Hafez:
Sorry, it's my first PR contributing to an open-source project, and you're right—somee of those files are generatd. When I cloned the repo, they were added automatically without me doing anything. But in the future, I'll make sure not to include them.

As for the "test" commit, I was just testing if it would upload to the repo. As I said, it's my first contribution to open source, so I apologize for that. I'll be more careful next time. Thanks for the guidance!

Issue #232, opened 2025-03-02
Avoid onboarding on import backup


lucaantonelli:
When we import a backup file in the app, it restart, but we should avoid it showing again the onboarding, since there should be already accounts and categories set by the imported file.

VoidElle:
@theperu I can work on this if you want

PR #342, opened 2025-03-19
fix: 232 | Rework of 'is_first_login' variable

VoidElle:
### What does this PR do?  
This PR resolves issue #232, where users who performed an import process were required to redo the onboarding process. Additionally, I noticed that the handling of the `is_first_login` variable in the main function might be not ideal.  

### Why was this bug occurring?  
The issue came from Phoenix, the library used to restart/reload the app after completing an import process. Phoenix re-executes the `build()` method of the `Launcher` widget, preventing the `_isFirstLogin` variable in `main.dart` from being properly reassigned. Its value was only retrieved inside the `main()` function, leading to incorrect behavior.  

### More bugs than expected  
While investigating, I also found another bug: when a user opens the app for the first time, the onboarding process is displayed. However, if they close the app while still in the onboarding flow, they will never see the onboarding process again. This happens because, in the current implementation, the `is_first_login` flag is set to `true` when the app executes the `main()` function (The first start does not get affected). This PR also addresses this issue by ensuring the flag is set to `true` only after the user completes onboarding.  

### Code changes  
To properly track whether the user has completed onboarding, the `is_first_login` value is now checked inside the `build()` method of the `Launcher` widget. Instead of storing its value in the `_isFirstLogin` variable (which was only retrieved in `main()`), the app now directly reads the updated value from `SharedPreferences` whenever `Launcher` is rebuilt.  

This ensures that whenever the app restarts, *whether due to Phoenix or any other reasons*, it always retrieves the latest `is_first_login` value, preventing incorrect onboarding behavior. The key name remains unchanged to avoid forcing existing users to redo the onboarding.  

### (!) Edit
A KV migration has been implemented, making users migrate from the `is_first_login` key to `onboarding_completed` key, without problems.

https://github.com/user-attachments/assets/791cd601-2052-450e-b754-8991fd3dbd08

https://github.com/user-attachments/assets/7fb4c3b9-f941-4f08-93e4-f6feb8e330eb

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -12,7 +12,7 @@ import 'routes.dart';
 import 'utils/app_theme.dart';
 import 'utils/notifications_service.dart';
 
-late bool _isFirstLogin;
+late SharedPreferences _sharedPreferences;
 
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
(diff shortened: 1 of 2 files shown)


lib/main.dart
```diff
@@ -46,14 +42,20 @@ class Launcher extends ConsumerWidget {
   // This widget is the root of your application.
   @override
   Widget build(BuildContext context, WidgetRef ref) {
+
+    // Although the saved key refers to 'is_first_login',
+    // its behavior aligns with a 'has the user completed onboarding' check.
+    // It has not been renamed to maintain compatibility with previous versions.
+    final bool isFirstLogin = _sharedPreferences.getBool('is_first_login') ?? true;
+
```
fres-sudo, line 50: Can't you provide a "migration" of the key? 

I mean a function that gets triggered in the`main()`, if the user has the `is_first_login` value set (only `false` will be possible after this change), then the value migrate also to the new `onboarding_completed` (or whatever) value.

I know it's a small detail, but it will get the job easier for the next dev that will see this.

VoidElle: Well, that's a really good idea to be honest.
I really did not like having the `is_first_login` key for handling a behaviour that was not the one auto-described in the name.

Thanks for the idea :)

VoidElle: Done! :D

PR #343, opened 2025-03-19
Ensure keyboard appears when adding/editing a budget

real-Hafez:
This PR resolves issue #337  , where the keyboard was not appearing properly when adding or editing a budget. The fix involves using AnimatedPadding with MediaQuery.of(context).viewInsets.bottom to ensure the UI adjusts dynamically when the keyboard appears.

  **_Additional Fixes & Improvements:_**

- Deleted a previous pull request #339  that contained unnecessary generated files (.lock, macOS-related files) and created this clean PR with only the required changes.
-  Cleaned up the commit history to include only meaningful changes.

### DIFF
--- a/lib/pages/planning_page/manage_budget_page.dart
+++ b/lib/pages/planning_page/manage_budget_page.dart
@@ -51,123 +51,131 @@ class _ManageBudgetPageState extends ConsumerState<ManageBudgetPage> {
 
   @override
   Widget build(BuildContext context) {
-    return Column(
-      children: [
-        Padding(
-          padding: const EdgeInsets.all(15),
-          child: Text(
-            "Select the categories to create your budget",
-            style: Theme.of(context).textTheme.titleLarge,
-          ),
-        ),
-        const Padding(
-          padding: EdgeInsets.symmetric(horizontal: 16, vertical: 8),
-          child: Row(
-            mainAxisAlignment: MainAxisAlignment.spaceBetween,
-            children: [
-              Text(
-                'CATEGORY',
-                textAlign: TextAlign.left,
-              ),
-              Text(
-                'AMOUNT',
-                textAlign: TextAlign.right,
-              ),
-            ],
-          ),
-        ),
-        Expanded(
-          child: Column(
-            children: [
-              ListView.builder(
-                shrinkWrap: true,
-                itemCount: budgets.length,
-                itemBuilder: (context, index) {
-                  return Dismissible(
-                    key: Key(budgets[index].idCategory.toString()),
-                    background: Container(
-                      padding: const EdgeInsets.only(right: 20.0),
-                      alignment: Alignment.centerRight,
-                      color: Colors.red,
-                      child: const Text(
-                        'Delete',
-                        textAlign: TextAlign.right,
-                        style: TextStyle(color: Colors.white),
-                      ),
-                    ),
-                    onDismissed: (direction) {
-                      deleteBudget(budgets[index], index);
-                    },
-                    child: BudgetCategorySelector(
-                      categories: categories,
-                      categoriesAlreadyUsed: categories
-                          .where((element) => budgets.map((e) => e.name).contains(element.name))
-                          .map((e) => e.name)
-                          .toList(),
-                      budget: budgets[index],
-                      initSelectedCategory: categories
-                              .where((element) => element.id == budgets[index].idCategory)
-                              .isEmpty
-                          ? categories[0]
-                          : categories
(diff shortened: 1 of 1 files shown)


bongio94:
Hello @real-Hafez 👋 

I would suggest to use `MediaQuery.viewInsetsOf(context).bottom` instead of `MediaQuery.of(context).viewInsets.bottom` to avoid rebuilding the descendant widget tree every time something from `MediaQuery` changes.

From the `MediaQuery`'s `.of()` method [documentation](https://api.flutter.dev/flutter/widgets/MediaQuery/of.html):

> Since it is typical that the widget only requires a subset of properties of the [MediaQueryData] object, prefer using the more specific methods (for example: [MediaQuery.sizeOf] and [MediaQuery.paddingOf]), as those methods will not cause a widget to rebuild when unrelated properties are updated.


Also, remember the trailing comma after `.bottom` for proper formatting 🙂

real-Hafez:
Hi @bongio94 👋
That makes sense! Before proceeding, let's hear @federicopozzato  thoughts on whether we should shift the button above the keyboard or handle it differently for iOS. I'm waiting for both of you to tell me the best approach to take. 😊

federicopozzato:
Yes, I opened an isse #244. I think those are different problems, so I suggest to handle them in different PR.
Otherwise, if you think that the problems are related, please specify better the problem you are solving with a screenshot (since is missing in the issue #337 )

Thanks!

real-Hafez:
Hi @federicopozzato  ,

I agree with you—both issues address the same problem but in different parts of the app. Issue #337  occurs in the budget section, while issue #244  affects the expenses section.

bongio94:
The issue is that iOS users won't be able to dismiss the keyboard if they don't have that floating button above it.

I think a possible solution would be to remove the focus from the `TextField` when tapping outside of it, so that the button can stay hidden by the keyboard and iOS users can still dismiss it by tapping anywhere else on the screen

federicopozzato:
@bongio94 Agreed. I think that would be also good to add a "Done" button in the keyboard (You know, the blue button that appears on the bottom righ side of the keyboard).

See attachement:
Screenshot 2025-03-21 at 9 33 06 AM-1
Screenshot 2025-03-21 at 9 33 03 AM

Here the example is with the search function in Whastapp. There are both states disabled and active.

bongio94:
I've looking into this yesterday and it seems like with Flutter, we can't show the _Done_ button within the numpad, only in the full size keyboard (the one you're showing in your screenshots)

federicopozzato:
So let's do what you said

> so that the button can stay hidden by the keyboard and iOS users can still dismiss it by tapping anywhere else on the screen

bongio94:
@real-Hafez Please let me know if you have any doubts on the proposed changes :) 

> I think a possible solution would be to remove the focus from the TextField when tapping outside of it, so that the button can stay hidden by the keyboard and iOS users can still dismiss it by tapping anywhere else on the screen

bongio94:
Hey @real-Hafez 

Yes, that should be everything 👍

real-Hafez:
✅ Fixed keyboard dismissal when tapping outside  
✅ Optimized MediaQuery usage to prevent unnecessary UI shifts.  
    Ready for review @bongio94! 🚀

real-Hafez:
Hi @napitek 

Sorry, I don’t have the time to check the conflicts at the moment. This task has been pending for more than a month, and I’ve tried reaching out to you on Discord, but it seems I wasn't able to get a response. Given the situation, I’ve decided not to continue working on this project.

mikev-cw:
@real-Hafez 
We’re all individual contributors who work on this project when we have time. We earn a living from other jobs.
What feels like a long wait for you, like a month, might not be the same for us. We don’t work for you or for this project full-time, and we’re sorry if that’s disappointing.

That said, just as we respect your choice to leave, we ask that you respect our time in return.
You called us at night, requested instant reviews, messaged all maintainers privately, and asked our community for personal advice, and we've always answered back. Always.

NOW is the time we were able to review this contribution. We're sorry if it felt too slow, but this is the pace we can maintain.
We hope to see you again in the future. 
Thank you for your time, and all the best.

real-Hafez:
Hi @mikev-cw 
I respect that this is a volunteer project — but respect goes both ways.
My messages weren’t demands, just efforts to align with your workflow and coding standards.
Reaching out at night your time was daytime and work hours in mine.
After a month of silence, it felt less like delay and more like disregard.
I’ve stayed professional throughout — I just wish that was mutual.
Wishing you the best going forward.

Issue #199, opened 2025-01-28
Bug: Recurring transaction should be also for Income and Transfers


theperu:
# 🐛 Bug Report

## 📋 Description

**Summary**:  
Recurring transaction should be also for Income and Transfers

**Expected Behavior**:  
The user should be able to select if a recurring transaction is for an income, expense or transfer transaction.

**Actual Behavior**:  
The recurring transaction is only for expense transactions.

Ic3b3rg:
Hi, if the issue is still available, I'd be happy to take it on. I believe I can implement the functionality to allow recurring transactions to be categorized as income, expense, or transfer. Let me know if that works for you!

PR #349, opened 2025-03-20
feat(transactions): enable recurring transactions for Income and Transfers

Ic3b3rg:
Recurring transactions were previously limited to Expenses.
This update allows recurring transactions to be used for Income and Transfers as well.

I noticed a problem, albeit non-blocking: the color of the selected transaction remains persistent even after the type change via the top button. This happens because the fallback value of the type is determined by the parameter passed to the function.

I chose not to intervene in this aspect because, given the current state management model, it would be necessary to rethink the assignment of the transaction type to the provider. The goal would be to ensure consistent management in line with the Single Source of Truth principle, avoiding duplication or inconsistent state.

#199

### DIFF
--- a/lib/model/recurring_transaction.dart
+++ b/lib/model/recurring_transaction.dart
@@ -10,6 +10,7 @@ class RecurringTransactionFields extends BaseEntityFields {
   static String toDate = 'toDate';
   static String amount = 'amount';
   static String note = 'note';
+  static String type = 'type';
   static String recurrency = 'recurrency';
   static String idCategory = 'idCategory';
   static String idBankAccount = 'idBankAccount';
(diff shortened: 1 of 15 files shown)


lib/custom_widgets/budget_circular_indicator.dart
```diff
@@ -39,17 +40,16 @@ class BudgetCircularIndicator extends ConsumerWidget with Functions {
                   children: [
                     TextSpan(
                       text: numToCurrency(amount),
-                      style: Theme.of(context)
-                          .textTheme
-                          .titleMedium!
-                          .copyWith(color: Theme.of(context).colorScheme.primary),
+                      style: Theme.of(context).textTheme.titleMedium!.copyWith(
+                          color: Theme.of(context).colorScheme.primary),
```
marcoredz, line 44: when accessing an InheritedWidget (like Theme) multiple times in a code block it would be better to declare it at the top of the `build` method (or inside the nearest Builder scope) in this way:

`final theme = Theme.of(context);`

then use just `theme` to access the properties:

Example:
`color: theme.colorScheme.primary`

lib/custom_widgets/budget_circular_indicator.dart
```diff
@@ -73,7 +73,9 @@ class BudgetCircularIndicator extends ConsumerWidget with Functions {
         ),
         const SizedBox(height: 10),
         Row(children: [
-          perc >= 0.9 ? const Icon(Icons.error_outline, color: Colors.red, size: 15) : Container(),
+          perc >= 0.9
+              ? const Icon(Icons.error_outline, color: red, size: 15)
+              : Container(),
```
marcoredz, line 78: When we want to show 'nothing' it would be better to use a `SizedBox()` widget instead of `Container` just because `Container` doesn't have a `const` constructor so we cannot indicate the keyword `const` before it.

`const` keyword should be placed everywhere when possible

Example:
`const SizedBox()` (or const `SizedBox.shrink()` to take up as little space as possible or `SizedBox.expand()` to take up as much space as possible) instead of just `Container()`.

lib/model/recurring_transaction.dart
```diff
@@ -144,23 +140,25 @@ class RecurringTransaction extends BaseEntity {
...
   Future<RecurringTransaction?> insert(RecurringTransaction item) async {
-    try{
+    try {
       final db = await database;
       final id = await db.insert(recurringTransactionTable, item.toJson());
       return item.copy(id: id);
-    } catch (e){
-      print(e);
+    } catch (e) {
+      print('insert ${e}');
```
marcoredz, line 161: do we want to use `print` here? `print` function is executed in production and normally we don't want to.

lib/model/recurring_transaction.dart
```diff
@@ -239,21 +237,23 @@ class RecurringTransactionMethods extends SossoldiDatabase {
...
       try {
-        lastTransactionDate = await _getLastRecurringTransactionInsertion(transaction.id ?? 0);
+        lastTransactionDate =
+            await _getLastRecurringTransactionInsertion(transaction.id ?? 0);
       } catch (e) {
         lastTransactionDate = transaction.fromDate;
       }
 
-      String entity = recurrenciesMap[transaction.recurrency]?['entity'] ?? 'UNMAPPED';
+      String entity =
```
marcoredz, line 246: I think this variable should be `final`

lib/model/recurring_transaction.dart
```diff
@@ -311,39 +315,44 @@ class RecurringTransactionMethods extends SossoldiDatabase {
...
       switch (scope) {
         case 'days':
-          lastTransactionDate = DateTime(lastTransactionDate.year, lastTransactionDate.month, lastTransactionDate.day + amount);
+          lastTransactionDate = DateTime(lastTransactionDate.year,
+              lastTransactionDate.month, lastTransactionDate.day + amount);
           break;
         case 'months':
           // get the last day of the next period
-          int lastDayOfNextPeriod = DateTime(lastTransactionDate.year, (lastTransactionDate.month + amount + 1), 0).day;
+          int lastDayOfNextPeriod = DateTime(lastTransactionDate.year,
```
marcoredz, line 323: also this variable should be `final`

lib/pages/planning_page/widget/recurring_payment_card.dart
```diff
@@ -40,21 +40,20 @@ class RecurringPaymentCard extends ConsumerWidget with Functions {
     final isDarkMode = ref.watch(appThemeStateNotifier).isDarkModeEnabled;
     final currencyState = ref.watch(currencyStateNotifier);
 
-    var cat = categories
+    var category = categories
```
marcoredz, line 43: `category `should be `final`

lib/pages/planning_page/widget/recurring_payment_card.dart
```diff
@@ -102,14 +101,22 @@ class RecurringPaymentCard extends ConsumerWidget with Functions {
                           children: [
                             Text("In ${getNextText()} days"),
                             const SizedBox(height: 10),
-                            Text(
-                                "-${transaction.amount}${currencyState.selectedCurrency.symbol}",
-                                style: const TextStyle(color: Colors.red)),
+                            Builder(builder: (context) {
+                              String prefix = prefixAmount(transaction.type);
```
marcoredz, line 105: `prefix `and `amountColor `should be final

lib/providers/transactions_provider.dart
```diff
@@ -197,15 +215,17 @@ class AsyncTransactionsNotifier extends AutoDisposeAsyncNotifier<List<Transactio
...
-    RecurringTransaction transaction = ref.read(selectedRecurringTransactionUpdateProvider)!.copy(
-        fromDate: ref.read(dateProvider),
-        toDate: ref.read(endDateProvider),
-        recurrency: recurrency.name.toUpperCase(),
-        amount: amount,
-        note: label,
-        idBankAccount: bankAccount.id!,
-        idCategory: category?.id,
-        updatedAt: DateTime.now());
+    RecurringTransaction transaction = ref
```
marcoredz, line 218: should be `final`

marcoredz (changes requested):

lib/model/recurring_transaction.dart
Ic3b3rg: yes i know. We'll need to implement something to handle error. We'll discuss it tomorrow

Ic3b3rg:
Thanks for your review <3

napitek:
@Ic3b3rg Hi, could you check the conflicts when you have time so we merge it?

PR #333, opened 2025-03-18
Resolve #292 - Enhance theme management with system theme support

xDefcon:
In reference to issue #292

Could someone please assist with testing this on iOS? I'm unable to test today.

It could be appropriate to involve an approval from the UI/UX team for this specific implementation.

PR #360, opened 2025-03-21
chore: Resolve #356 - Reworked currency selector in General Settings

### LINKED ISSUE #356: [Bug] Select a currency dialog has two background colors
### 🛰️ **What happened?**

When selecting a currency from:

_Settings_ -> _General Settings_ -> Tap on _currency_ symbol

The dialog that pops up for selecting the currency shows two different background colors, one for the dialog itself and one for the `ListTile`s.

### 🖥️ **OS**

Both

### 🔄 **What version?**

Android 15 

### 🔢 **Version**

0.1.2

### 📸 **Screenshots or logs**

https://github.com/user-attachments/assets/d3c733ab-07ed-4a78-931f-5f2378a1da6b

https://github.com/user-attachments/assets/215ba9b8-15de-4e63-b357-542e8ad17d5a

### 📝 **Code of Conduct**

- [x] ✅ I agree

bongio94:
## 🎯 Description

This PR reworks the currency selector that can be found in _Settings_ -> _General Settings_.
The old currency selector had some issues with the colors of the` ListTile`s and, taking in consideration the discussion that is happening in #333 about adding the system theme as option, I went with the updated design.

Design follows [this](https://www.figma.com/design/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?node-id=3286-9189&t=5wYGZcAdXnVsZBaa-11) Figma frame.

Diving a bit deeper in the changes:

- Added `onSurfaceVariant` to `customColorScheme` and `darkCustomColorScheme` to have a lighter color that gives enough contrast when used on `surface` (`darkBlue6`). Please check these messages ([[1]](https://github.com/RIP-Comm/sossoldi/issues/356#issuecomment-2743009560), [[2]](https://github.com/RIP-Comm/sossoldi/issues/356#issuecomment-2743037750)) for more details on this addition.

- In `lib\pages\general_options\widgets` I've added the `selector` folder. It contains two files, `selector_container.dart` and `selector_tile.dart`:
  - `SelectorContainer` is a simple `Widget` responsible for the background and the top label. It takes a `label` and a `child`.
  - `SelectorTile` is the list tile like UI element which shows the currencies list (and ideally should also be used to show the appearance options).  
   I've tried to make it flexible while still maintaining the convenience of use. The important part for me here was to have a widget that can give a uniform look to the settings page and that can be reused when more settings will be added to the page. I know it will need to be updated as requirements will change but I can take care of that when time comes.  
   It has a customisable `title`, `trailing` (_String_) and `trailingWidget`. This widget is meant to handle selection between a number of options, for this purpose the `isSelected` _bool_ can be used. Also added a _void_ `onTap`.
- The old `CurrencySelectorDialog` has been renamed to `SettingsCurrencySelector` and it uses `SelectorContainer` and `SelectorTile` while maintaining the same currency selection state handling that was used in the previous implementation.
- `SettingsCurrencySelector` is then used in `lib\pages\general_options\general_settings.dart`.
- For what concern testing:  
  - Added some simple tests for the `SelectorTile` basic interactions.
  - Added a convenience `extension` on `WidgetTester` in `test\test_utils`. It's called `pumpWidgetWithMaterialApp` and as the name suggests, it's just a wrapper on `pumpWidget` which includes a `MaterialApp` and the `Material` widget. I've added an option to toggle on or off the `ProviderScope` as well (defaults to on).
  

Closes: #356   

## 📱 Changes

- [X] Describe key changes made
- [X] List any new features, bug fixes, or refactors
- [X] Include additional details if necessary

### Behaviour
1. Go to _Settings_ -> _General Settings_
2. Change your selected currency

| Before | After |
|----------|----------|
|https://github.com/user-attachments/assets/e949b3cc-c240-46d1-a46f-5a9b345f1f3d | https://github.com/user-attachments/assets/b4ac15cb-11c7-43df-9047-16cc669f5645 |

## 🔍 Checklist for reviewers
- [X] Code is formatted correctly
- [X] Tests are passing
- [X] New tests are added (if needed)
- [X] Style matches the figma/designer requests
- Tested on:
    - [ ] iOS
    - [X] Android

## ✍️ Additional Context

Note that this new UI might look a bit out of place because the appearance switcher is still the same. Hopefully when both components use this new layout it'll look better as a whole 😅

### DIFF
--- a/lib/pages/general_options/general_settings.dart
+++ b/lib/pages/general_options/general_settings.dart
@@ -3,7 +3,6 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 
 import '../../constants/style.dart';
 import '../../model/currency.dart';
-import '../../providers/currency_provider.dart';
 import '../../providers/theme_provider.dart';
 import 'widgets/currency_selector.dart';
 
@@ -32,7 +31,6 @@ class _GeneralSettingsPageState extends ConsumerState<GeneralSettingsPage> {
   @override
   Widget build(BuildContext context) {
     final appThemeState = ref.watch(appThemeStateNotifier);
-    final currencyState = ref.watch(currencyStateNotifier);
 
     return Scaffold(
       appBar: AppBar(
@@ -77,32 +75,7 @@ class _GeneralSettingsPageState extends ConsumerState<GeneralSettingsPage> {
               ],
             ),
             const SizedBox(height: 20),
-            Row(
-              children: [
-                Text("Currency",
-                    style: Theme.of(context)
-                        .textTheme
-                        .titleLarge!
-                        .copyWith(color: Theme.of(context).colorScheme.primary)),
-                const Spacer(),
-                GestureDetector(
-                    onTap: () {
-                      setState(() {
-                        CurrencySelectorDialog.selectCurrencyDialog(
-                            context, currencyState, currencyList);
-                      });
-                    },
-                    child: CircleAvatar(
-                        radius: 30.0,
-                        backgroundColor: blue5,
-                        child: Center(
-                            child: Text(
-                          currencyState.selectedCurrency.symbol,
-                          style: TextStyle(
-                              color: Theme.of(context).colorScheme.onPrimary, fontSize: 25),
-                        )))),
-              ],
-            ),
+            SettingsCurrencySelector(),
             const SizedBox(height: 20),
             /*
             Row(
--- a/lib/pages/general_options/widgets/currency_selector.dart
+++ b/lib/pages/general_options/widgets/currency_selector.dart
@@ -1,76 +1,71 @@
 import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:gap/gap.dart';
 
-import '../../../constants/style.dart';
 import '../../../model/currency.dart';
 import '../../../providers/currency_provider.dart';
+import 'selector/selector_container.dart';
+import 'selector/selector_tile.dart';
 
-class CurrencySelectorDialog {
-  static void selectCurrencyDialog(
-      BuildContext context, CurrencyState state, Future<List<Currency>> currencies) {
-    showDialog(
-      context: context,
-      builder: (context) => AlertDialog(
-        title: Text(
-          'Select a currency',
-          style: Theme.of(context)
-              .textTheme
-              .titleLarge!
-              .copyWith(color: Theme.of(context).colorScheme.primary),
-        ),
-        content: SizedBox(
-          height: 300,
-          width: 220,
-          child: SingleChildScrollView(
-            child: FutureBuilder(
-                future: currencies,
-                builder: (context, snapshot) {
-                  if (snapshot.hasData &&
-                      snapshot.data != null &&
-                      snapshot.connectionState == ConnectionState.done) {
-                    return ListView.builder(
-                      itemCount: snapshot.data!.length,
-                      physics: const NeverScrollableScrollPhysics(),
-                      shrinkWrap: true,
-                      itemBuilder: (BuildContext context, int i) {
-                        return GestureDetector(
-                          onTap: () {
-                            state.setSelectedCurrency(snapshot.data![i]);
-                            Navigator.pop(context);
-                          },
-                          child: ListTile(
-                            leading: CircleAvatar(
-                              radius: 22,
-                              backgroundColor: state.selectedCurrency.code == snapshot.data![i].code
-                                  ? blue5
-                                  : grey1,
-                              child: Text(snapshot.data![i].symbol,
-                                  style: TextStyle(
-                                      color: Theme.of(context).colorScheme.onPrimary,
-                                      fontSize: 20)),
-                            ),
-                            title: Text(
-                              snapshot.data![i].name,
-                              textAlign: TextAlign.center,
-                            ),
-                          ),
-                        );
-                      },
-                    );
-                  } else if (snapshot.hasError) {
-                    return Text('Something went wrong: ${snapshot.error}');
-                  } else {
-                    if (snapshot.connectionState == ConnectionState.waiting) {
-                      return Transform.scale(
-                        scale: 0.5,
-                        child: const CircularProgressIndicator(),
-                      );
-                    } else {
-                      return const Text("Search for a transaction");
-                    }
-                  }
-                }),
-          ),
-        ),
+class SettingsCurrencySelector extends ConsumerStatefulWidget {
+  const SettingsCurrencySelector({super.key});
+
+  @override
+  ConsumerState<ConsumerStatefulWidget> createState() =>
+      _SettingsCurrencySelectorState();
+}
+
+class _SettingsCurrencySelectorState
+    extends ConsumerState<SettingsCurrencySelector> {
+  final currenciesFuture = CurrencyMethods().selectAll();
+
+  @override
+  Widget build(BuildContext context) {
+    final currencyState = ref.watch(currencyStateNotifier);
+
+    return SelectorContainer(
+      label: 'SELECT A CURRENCY',
+      child: FutureBuilder(
+        future: currenciesFuture,
+        builder: (context, snapshot) {
+          if (snapshot.hasData &&
+              snapshot.data != null &&
+              snapshot.connectionState == ConnectionState.done) {
+            final currencies = snapshot.data;
+
+            return ListView.separated(
+              separatorBuilder: (context, index) {
+                return Gap(8);
+              },
+              shrinkWrap: true,
+              itemCount: currencies!.length,
+              itemBuilder: (context, index) {
+                final selected = currencyState.selectedCurrency.code ==
+                    currencies[index].code;
+
+                return SelectorTile(
+                  onTap: () {
+                    currencyState.setSelectedCurrency(currencies[index]);
+                  },
+                  title: currencies[index].name,
+                  trailing: currencies[index].symbol,
+                  isSelected: selected,
+                );
+              },
+            );
+          } else if (snapshot.hasError) {
+            return Text('Something went wrong: ${snapshot.error}');
+          } else {
+            if (snapshot.connectionState == ConnectionState.waiting) {
+              return Transform.scale(
+                scale: 0.5,
+                child: const CircularProgressIndicator(),
+              );
+            } else {
+              return const Text("Search for a transaction");
+            }
+          }
+        },
       ),
     );
   }
--- a/lib/pages/general_options/widgets/selector/selector_container.dart
+++ b/lib/pages/general_options/widgets/selector/selector_container.dart
@@ -1,0 +1,40 @@
+import 'package:flutter/material.dart';
+import 'package:gap/gap.dart';
+
+class SelectorContainer extends StatelessWidget {
+  final String? label;
+  final Widget? child;
+
+  const SelectorContainer({
+    this.label,
+    this.child,
+    super.key,
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    final appTheme = Theme.of(context);
+
+    return Container(
+      padding: const EdgeInsets.all(12),
+      decoration: BoxDecoration(
+        borderRadius: BorderRadius.circular(4),
+        color: appTheme.colorScheme.surface,
+      ),
+      child: Column(
+        mainAxisSize: MainAxisSize.min,
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          if (label != null && label!.isNotEmpty) ...[
+            Text(
+              label!.toUpperCase(),
+              style: appTheme.textTheme.labelMedium,
+            ),
+            Gap(8),
+          ],
+          child ?? const SizedBox.shrink(),
+        ],
+      ),
+    );
+  }
+}
--- a/lib/pages/general_options/widgets/selector/selector_tile.dart
+++ b/lib/pages/general_options/widgets/selector/selector_tile.dart
@@ -1,0 +1,69 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:gap/gap.dart';
+
+import '../../../../constants/style.dart';
+
+class SelectorTile extends ConsumerWidget {
+  final String title;
+  final String? trailing;
+  final Widget? trailingWidget;
+  final bool isSelected;
+  final VoidCallback? onTap;
+
+  const SelectorTile({
+    required this.title,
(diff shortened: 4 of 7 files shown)


fres-sudo:
I'll try to review this tomorrow evening or sunday, thanks for the PR!

bongio94:
Regarding this comment I made:

> Note that this new UI might look a bit out of place because the appearance switcher is still the same. Hopefully when both components use this new layout it'll look better as a whole 😅


I think this could be improved a bit further and make it ready for future settings by making the `SelectorContainer` an expandable tile. 

Here's a quick demo;


https://github.com/user-attachments/assets/a649a067-4e35-4c05-bce5-a1a430d83d15

fres-sudo:
> I think this could be improved a bit further and make it ready for future settings by making the `SelectorContainer` an expandable tile.

I think it's a very cool design the one in the demo, why not implement it? Since we actually introduced 3 theme modes in #333?

lib/pages/general_options/widgets/currency_selector.dart
```diff
@@ -1,76 +1,71 @@
...
+                  trailing: currencies[index].symbol,
+                  isSelected: selected,
+                );
+              },
+            );
+          } else if (snapshot.hasError) {
+            return Text('Something went wrong: ${snapshot.error}');
+          } else {
+            if (snapshot.connectionState == ConnectionState.waiting) {
+              return Transform.scale(
```
fres-sudo, line 60: Why a `Transform`? Why don't use build in `contstraint` property of the `CircularProgressIndicator`?

Or even butter would be to create a custom loader. I usually use something like that in my projects:

```dart

class LoadingSpinner extends StatelessWidget {
  const LoadingSpinner({super.key, this.size = 24, this.isAdaptive = true, this.color});

  final double size;
  final bool isAdaptive;
  final Color? color;

  @override
  Widget build(BuildContext context) => Center(
        child: SizedBox(
          height: size,
          width: size,
          child: isAdaptive
              ? CircularProgressIndicator.adaptive(
                  backgroundColor: color ?? Theme.of(context).colorScheme.primary,
                )
              : CircularProgressIndicator(
                  valueColor: AlwaysStoppedAnimation<Color>(color ?? Theme.of(context).colorScheme.primary),
                ),
        ),
      );
}
```

would be cool to see something similar here.

lib/pages/general_options/widgets/selector/selector_container.dart
```diff
@@ -0,0 +1,63 @@
...
+      child: Column(
+        mainAxisSize: MainAxisSize.min,
+        crossAxisAlignment: CrossAxisAlignment.start,
+        children: [
+          if (label != null && label!.isNotEmpty) ...[
+            Text(
+              label!.toUpperCase(),
+              style: appTheme.textTheme.labelMedium,
+            ),
+            Gap(8),
```
fres-sudo, line 56: Again, Gap must be avoided

lib/pages/general_options/widgets/selector/selector_tile.dart
```diff
@@ -0,0 +1,88 @@
...
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final appTheme = Theme.of(context);
+
+    return InkWell(
+      onTap: onTap,
+      child: Container(
+        padding: const EdgeInsets.all(16),
+        foregroundDecoration: BoxDecoration(
+          borderRadius: BorderRadius.circular(4),
```
fres-sudo, line 51: Is this the default `BorderRadius` that we're using? Isn't it 8?

P.S. Would be great to see here a `constant.dart` file to be more consistent and to avoid errors by new contributors.

lib/pages/general_options/widgets/selector/selector_tile.dart
```diff
@@ -0,0 +1,88 @@
...
+                title,
+                style: appTheme.textTheme.bodyMedium?.copyWith(
+                  fontWeight: isSelected ? FontWeight.w700 : FontWeight.w400,
+                  color: isSelected
+                      ? appTheme.colorScheme.onSurfaceVariant
+                      : appTheme.colorScheme.onSurface,
+                ),
+              ),
+            ),
+            Gap(8.0),
```
fres-sudo, line 70: Gap again

lib/pages/general_options/widgets/selector/selector_tile.dart
```diff
@@ -0,0 +1,88 @@
...
+class SelectorTile extends ConsumerWidget {
+  /// The title [String] of the tile, will be displayed on the left side of the tile.
+  final String title;
+
+  /// An optional trailing [String], will be displayed on the right side of the tile.
+  final String? trailing;
+
+  /// An optional trailing [Widget], will be displayed on the right side of the
+  /// tile only if [trailing] is `null`.
+  final Widget? trailingWidget;
```
fres-sudo, line 22: I think this approach is a bit confusing and error-prone. Why don't provide just a single:

```dart
final Widget? trailing;
```
If the problem is the annoying repetition of the styling of the text (if any), I think you can solve the problem just by wrap with `ThemeData` the `trailing` widget:

```dart
Theme(
   data: Theme.of(context).copyWith(...your custom styling (overridable)...)
   child: trailingWidget
)
```
I didn't test this solution yet, but at first glance it should work.

Ofc you have to pass also a `trailingTextStyle`.

Other than that, it you wanna stick with this approach, you should at least provide some assertions, to reinforce the concept: you should put an `assert` that says if `trailing != null` then `trailingWidget` should be `null` and vice versa.

fres-sudo (changes requested):
Thanks for the PR!

I tested on iOS, it works just fine, the only small issue that I found is that ambigous spacing:

simulator_screenshot_F31ADA43-0D28-4420-BAF9-5B28A3910AAA

You can easily spot it under the last element of the list...is it intended? bcuz I can't see it in your last video demo..

bongio94:
Hey @fres-sudo 

Thanks for the detailed review 🙌 
I'll start to go through your comments now 🙂

bongio94:
> > I think this could be improved a bit further and make it ready for future settings by making the `SelectorContainer` an expandable tile.
> 
> I think it's a very cool design the one in the demo, why not implement it? Since we actually introduced 3 theme modes in #333?

I would want to wait for @federicopozzato to approve this first. 
While I think it's in line with the Figma specs, it's still a bit different.

lib/pages/general_options/widgets/currency_selector.dart
```diff
@@ -1,76 +1,71 @@
...
+        future: currenciesFuture,
+        builder: (context, snapshot) {
+          if (snapshot.hasData &&
+              snapshot.data != null &&
+              snapshot.connectionState == ConnectionState.done) {
+            final currencies = snapshot.data;
+
+            return ListView.separated(
+              separatorBuilder: (context, index) {
+                return Gap(8);
```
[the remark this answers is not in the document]
bongio94: Yep, I'll get rid of all of the `Gap`s

lib/pages/general_options/widgets/currency_selector.dart
bongio94: This is a pre-existing piece of code, I didn't want to change it as part of this MR.

I agree on having a reusable loading indicator tho

lib/pages/general_options/widgets/selector/selector_tile.dart
bongio94: As @federicopozzato said, this is probably part of a bigger piece of work around the design system.

On the radius, that's taken from Figma

lib/pages/general_options/widgets/selector/selector_tile.dart
bongio94: Yeah, I thought about it, but having two separate properties was the most straight forward solution.

Being the `trailingWidget` a `Widget`, anything could go there, so how would you know which style to override in the `copyWith`?

I could override the text theme by default (so that when the `trailingWidget` is a `Text` it would use the correct styling for selected and unselected state) but if an Icon is passed, for example, that would apply of course, so, in any case, the style must be outside of the `SelectorTile` widget.

I can add the` assert`s tho

bongio94:
> Thanks for the PR!
> 
> I tested on iOS, it works just fine, the only small issue that I found is that ambigous spacing:
> 
> simulator_screenshot_F31ADA43-0D28-4420-BAF9-5B28A3910AAA
> 
> You can easily spot it under the last element of the list...is it intended? bcuz I can't see it in your last video demo..

hmm, I don't think that's the `separatorBuilder` from `ListView.separated`, as you can see it's not showing in the original screenshot attached.

Let's wait for the new layout to be pushed and you can re-test

lib/pages/general_options/widgets/selector/selector_tile.dart
fres-sudo: >Being the trailingWidget a Widget, anything could go there, so how would you know which style to override in the copyWith?

My intent was actually to replicate the `ListTile` behaviour that has a defualt text style, but you can override, I didn't dive too much in the actual `ListTile` implementation, but I think that this would make the job done in our case. I didn't try to implement it yet but seems promising. This is the original `ListTile` `leadingAndTrailingStyle` implementation :

```dart
    TextStyle? leadingAndTrailingStyle;
    if (leading != null || trailing != null) {
      leadingAndTrailingStyle =
          leadingAndTrailingTextStyle ??
          tileTheme.leadingAndTrailingTextStyle ??
          defaults.leadingAndTrailingTextStyle!;
      final Color? leadingAndTrailingTextColor = effectiveColor;
      leadingAndTrailingStyle = leadingAndTrailingStyle.copyWith(
        color: leadingAndTrailingTextColor,
      );
    }


    Widget? leadingIcon;
    if (leading != null) {
      leadingIcon = AnimatedDefaultTextStyle(
        style: leadingAndTrailingStyle!,
        duration: kThemeChangeDuration,
        child: leading!,
      );
    }
```
You can find this code [here](https://github.com/flutter/flutter/blob/d25349eb1c7a0dbae4f3a734d34f07ddda74adaf/packages/flutter/lib/src/material/list_tile.dart#L872C1-L891C6)

But don't spend more than 5min on this if it doesnt' work. The assert will make the job done tho

fres-sudo:
> Let's wait for the new layout to be pushed and you can re-test

Wdym by the "new layout" what did I miss?

bongio94:
> > Let's wait for the new layout to be pushed and you can re-test
> 
> Wdym by the "new layout" what did I miss?

Wrong word probably as the layout is the same 😅 
I meant the new interaction, with the expansion functionality

federicopozzato:
Guys, we appreciate you effort and propositivity, but it's always a metter of process. Things has to be DESIGNED FIRST, then developed. I can't stress this enough. 
Multicurrency is something we have to work on for sure, but I'm not sure that here will be just a currency selector as a feature: what if I want to have multiple currencies at the same time?  @theperu I don't remember what we said about it for the 1.0

theperu:
@federicopozzato this is not an implementation of the multi currency feature. This was opened to align the graphic of the theme selector that has changed in #333 so that we don't have two completely different styles between the two elements

bongio94:
@theperu @federicopozzato Should I move forward with this?
https://github.com/RIP-Comm/sossoldi/pull/360#issuecomment-2745141974

theperu:
Ok for me but I'll leave the last word to @federicopozzato

lib/pages/general_options/widgets/selector/selector_tile.dart
bongio94: @fres-sudo 
For now, I've just added an optional `TextStyle` that can override the default one.

We can revisit this Widget at a later date when we'll have more settings that can use it 👍 
How does that sounds?

bongio94:
PR details have changes a bit after the latest commits, I'll try to go through the latest changes with this comment and I'll link this to the PR description.

After @fres-sudo suggestions: 

- I've removed all of the `Gap` widget used (`gap` dependency has been removed).
- Added `assert`s on `SelectorTile` constructor for ensuring that only one between `trailing` and `trailingWidget` is used.  
  I've also added an optional `trailingTextStyle` which gives the ability of customizing the `trailing` text `TextStyle`. 
- Added the expansion behaviour. In particular:
  - `SelectorContainer` now has an `initiallyExpanded` property, which controls the initial expansion state (whether or not the container should be expanded on instantiation), and a `hasTrailingChevron` property which toggles on or off the trailing chevron which rotates when the container expands or collapses.
  - The `label` changes basing on the expansion status. When collapsed the label is bigger (fontSize 14, w600), when expanded is smaller (fontSize 12, w400)

bongio94:
@fres-sudo Could you have a look at my comment [here](https://github.com/RIP-Comm/sossoldi/pull/360#discussion_r2017591261) and close the change request if it sounds good?

lib/pages/general_options/widgets/selector/selector_tile.dart
fres-sudo: Sorry I completely missed that comment, yes I agree with you, the `trilingTextStyle` attribute is enough!

lib/pages/general_options/general_settings.dart
```diff
@@ -86,41 +84,9 @@ class _GeneralSettingsPageState extends ConsumerState<GeneralSettingsPage> {
...
-                          fontSize: 25,
-                        ),
-                      ),
-                    ),
-                  ),
-                ),
-              ],
-            ),
-            const SizedBox(height: Sizes.lg),
+            const SizedBox(height: 20),
```
fres-sudo, line 87: Here see what fits better:

```dart
const SizedBox(height: Sizes.lg), //that is 16
//or
const SizedBox(height: Sizes.xl), //that is 24
```

If none of them fits well, just go with 

```dart
const SizedBox(height: Sizes.lg + Sizes.xs),
```
or just leave it as it is nvm

lib/pages/general_options/general_settings.dart
```diff
@@ -86,41 +84,9 @@ class _GeneralSettingsPageState extends ConsumerState<GeneralSettingsPage> {
...
+            SettingsCurrencySelector(),
+            const SizedBox(height: 20),
```
fres-sudo, line 89: Same as above

fres-sudo (changes requested):
I saw that you correctly rebased everything, thx!!!
Just a small miss.

lib/pages/general_options/general_settings.dart
bongio94: Missed these, thanks.

Should be fixed now 👍

lib/pages/general_options/general_settings.dart
bongio94: Missed these, thanks.

Should be fixed now 👍

theperu:
Hi @bongio94👋

Thanks!

Issue #361, opened 2025-03-21
[REFACTORING] Remove Gap as dependency


fres-sudo:
### 📝 **Improvement Description**

[Gap](https://pub.dev/packages/gap) is at that point of Flutter (v3.27+) almost useless, since  [the introduction](https://docs.flutter.dev/release/release-notes/release-notes-3.27.0)  of the spacing property.

### 💡 **Proposed Solution**

Remove completely `Gap`as dep and refactor

### 🌿 **Benefits**

Less dep and cleaner native solution with `spacing`.

### 🧱 **Impact Area**

- [ ] Backend
- [x] Frontend
- [ ] Infrastructure
- [ ] DevOps
- [ ] Other

### ✍️ **Additional Context**

I can work on that

### 📜 **Code of Conduct**

- [x] ✅ I agree to the Code of Conduct

PR #370, opened 2025-03-22
del(pubspec): remove gap as dependency

fres-sudo:
## 🎯 Description

Gap was not being used in the whole project and so it was an unnecessary dependency

Closes: #361

## 📱 Changes

Removed `gap` from `pubspec.yml`

## 🔍 Checklist for reviewers
- [x] Tests are passing
- Tested on:
    - [x] iOS
    - [ ] Android

## ✍️ Additional Context

Don't think that Android tests are needed here 😄

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -22,7 +22,6 @@ dependencies:
   flutter_native_splash: ^2.4.4
   flutter_phoenix: ^1.1.1
   flutter_riverpod: ^2.6.1
-  gap: ^3.0.1
   intl: ^0.19.0
   path: ^1.9.1
   path_provider: ^2.1.5


pubspec.yaml
```diff
@@ -109,4 +108,4 @@ flutter_icons:
   image_path: "assets/icon.png"
   image_path_ios: "assets/iosicon.png"
   adaptive_icon_background: "#ffffff"
-  adaptive_icon_foreground: "assets/icon.png"
\ No newline at end of file
+  adaptive_icon_foreground: "assets/icon.png"
```
fres-sudo, line 111: The line here was a previous error of formatting, a new line at the end of the file should always be there

bongio94:
Only one minor thing from me: should we remove the `pubspec.lock`'s dependency specification as well?

Btw, I don't get why this dependency was added in the first place as it isn't used throughout the project 😅

fres-sudo:
> Only one minor thing from me: should we remove the `pubspec.lock`'s dependency specification as well?

Yep you're right, I miss that, I was too used to remove the lock file be default from the working tree, I'll fix it.

> Btw, I don't get why this dependency was added in the first place as it isn't used throughout the project 😅 

I don't know it either...

lucaantonelli:
> Btw, I don't get why this dependency was added in the first place as it isn't used throughout the project 😅

Actually, i think i added it by mistake, but i wanted to use it since i think it's clearer to use gap instead of SizedBox(height:20) or SizedBox(width:20) you can just write Gap(20), it simplify making spaces and it's not an expensive package, maybe during the refactor we can use it, let me know what you think @fres-sudo @bongio94

fres-sudo:
>  it simplify making spaces and it's not an expensive package

To me, specifying the `width`or `height` it's just a way to make the code clearer to other person.

Plus the [gap repo](https://github.com/letsar/gap) doesn't receive any updates for 2 years, I know that in this case it's not a big deal bcuz it's a simple widget, but it's another thing that makes me want to walk away from it.

One thing that we really need, instead of another widget, is a set of predefined sizes that we can use like that:

```dart
const SizedBox(height: Sizes.xl);
```

bongio94:
Tbh I don't have a strong opinion on this.

What I can say is that I usually specify my own spacing when working on projects.

The important thing for me when adding spaces is that the declaration should be as fast as possible, I don't wanna loose time declaring a space widget (`gap` was actually quite ok on this aspect).
The way I usually go is with `HSpace` and `VSpace` from which you call named constructors that represent size (t-shirt). So you end up with:

```dar
[
   Widget(),
   VSpace.m()
   Widget()
]
```

bongio94:
From one of my projects


```dart
class HSpace extends StatelessWidget {
  final double? width;
  const HSpace({
    super.key,
    this.width = 12.0,
  });

  const HSpace.xxs({super.key}) : width = 4.0;
  const HSpace.xs({super.key}) : width = 8.0;
  const HSpace.s({super.key}) : width = 12.0;
  const HSpace.m({super.key}) : width = 16.0;
  const HSpace.l({super.key}) : width = 32.0;
  const HSpace.xl({super.key}) : width = 64.0;

  @override
  Widget build(BuildContext context) {
    return SizedBox(
      width: width,
    );
  }
}

class VSpace extends StatelessWidget {
  final double? height;
  const VSpace({
    super.key,
    this.height = 12.0,
  });

  const VSpace.xxs({super.key}) : height = 4.0;
  const VSpace.xs({super.key}) : height = 8.0;
  const VSpace.s({super.key}) : height = 12.0;
  const VSpace.m({super.key}) : height = 16.0;
  const VSpace.l({super.key}) : height = 32.0;
  const VSpace.xl({super.key}) : height = 64.0;

  @override
  Widget build(BuildContext context) {
    return SizedBox(
      height: height,
    );
  }
}
```

fres-sudo:
I am cool with that option, the only thing that I dislike is this:

> you call named constructors that represent size (t-shirt)

I'd prefer to go with standard tailwind css breakpoint naming, such as: `xs`, `sm`, `md`, `lg`, `xl`,

P.S. `2xl` is not possible in dart since you cannot define a method that starts with a number, so I usually go with `xxl` and `xxs`

lucaantonelli:
Based on that, i'm going to accept the pr, we will discuss better that thing in the refactoring process. Thanks to both!

fres-sudo:
Lately I've been using this very handy set of functions:

https://github.com/fres-sudo/pocket_flutter/blob/master/lib/ui/device.dart

fres-sudo:
> Based on that, i'm going to accept the pr, we will discuss better that thing in the refactoring process. Thanks to both!

Damn, I didn't actually update the `pubspec.lock` file, not a big deal some other can push their updated `pubspec.lock`file to `main`

lucaantonelli:
> > Based on that, i'm going to accept the pr, we will discuss better that thing in the refactoring process. Thanks to both!
> 
> Damn, I didn't actually update the `pubspec.lock` file, not a big deal some other can push their updated `pubspec.lock`file to `main`

I'm going to update it. No problem :)

Issue #373, opened 2025-03-23
[Tech Improvement] - Add a set of utilities for spacing, sizes and responsiveness.


fres-sudo:
### 📝 **Improvement Description**

To speed up the refactoring process, I think it can be appropriate to start working at least on those small stuffs.

So I think it will be appropriate to add a set of `class` utilities to allow every contributor to stay on the same track with the design system and ensure consistency across all the app.

### 🛑 **Problem Statement**

So far, there's not any kind of guidelines about the sizing of the UI element.

### 💡 **Proposed Solution**

Start adding a `ui/` folder, as proposed in [this discussion](https://github.com/RIP-Comm/sossoldi/discussions/351).

Under the `ui/` folder, add a `device.dart` file that contains [this set of utilities](https://github.com/fres-sudo/pocket_flutter/blob/master/lib/ui/device.dart).

Ofc this will lead to a complete refactor across all the app that I can manage.

This `device.dart` contains only utilities about constants and responsive designs (that I think are needed for the app being cross-platform, also for desktop and tablet).

As discussed in #370 we can introduce some other utilities for spacing, if everyone agrees, we can introduce also a `sizing.dart` for the sizing utilities.

### 🌿 **Benefits**

Better DX and consistent design system.

### 🧱 **Impact Area**

- [ ] Backend
- [x] Frontend
- [ ] Infrastructure
- [ ] DevOps
- [ ] Other

### 📜 **Code of Conduct**

- [x] ✅ I agree to the Code of Conduct

fres-sudo:
If @lucaantonelli and @bongio94 don't have anything to add, I can start work on a PR for that issue.

I also wait for @federicopozzato hints about the actual sizing (won't be a problem to tweak also during the PR review, it will be just a number change).

federicopozzato:
@fres-sudo This is a good thing for sure, thanks for pointing it out!

It’s in our design backlog.
This is a big thing, to be addressed from our side first. 
We as a design team will have to create those guidelines (like a kind of design system/UI kit). 

This will of course take some time, so please push this issue later in the development phase.

So far in the figma all the main cases and screens are displayed, so you just have to stick to those.

bongio94:
While I agree with @federicopozzato, I think we can still create a spacing system that's consistent and convenient to use as discussed in #370.

When the design system will be in a good place, we could start creating the same convenience structures around everything else (ie: border radius, border width, shadows, colors, animations durations and curves)

PR #378, opened 2025-03-25
style: added and refactored styling utilities

fres-sudo:
## 🎯 Description

This PR tries to refactor all the inconsistency of the styling across the code base.

In particular, those properties:

- Responsiveness
- Assets
- Border Radius
- Edge Insets
- Spacing (SizedBox)

weren't globally handled, leading to a big inconsistency, and hard time for new devs to approach the project.

On top of that, this PR introduces also the `extension` concept globally, to prefer over mixin for our use case.

This PR also move the position of certain widgets under the new `ui/` folder.

Other small improvements and refactoring are made here and there.

Closes: #373 

## 🔍 Checklist for reviewers
- [x] Tests are passing
- Tested on:
    - [x] iOS
    - [x] Android


lib/ui/assets.dart
napitek, line 1: here I think the `images` folder is missing, since the images are all in `assets`

lib/ui/device.dart
napitek, line 1: What happens on foldable phones? 
I would add `foldable` in the enum, and then handle the logic to detect changes in the state of the foldable device. 
https://api.flutter.dev/flutter/widgets/MediaQueryData/displayFeatures.html

lib/ui/assets.dart
fres-sudo: You're right! I actually created a new `assets/images/` folder under in order to have the assets used in the flutter app in a separated folder.
By doing that, the "mapping" inside `assets.dart` is correct,

lib/ui/device.dart
fres-sudo: I understand the point here, but I honestly don't understand what I am supposed to do on foldable devices? I'd never encountered this situation (+ I never used a foldable device so far), so I am a bit lost. 

So what we're supposed to do when the device is foldable?

lib/ui/assets.dart
```diff
@@ -0,0 +1,6 @@
+/// Assets class to manage all the assets in the app
+///
+abstract final class SossoldiAssets {
+  static String get openVault => "assets/images/logo.png";
```
lucaantonelli, line 4: This should be:
openVault => "assets/images/open_vault.png";

fres-sudo: damn I completely miss that, my bad

lucaantonelli: No problem, but still, to use assets in subdirectory, you need to specify the folder in the pubspec, like stated here:
https://docs.flutter.dev/ui/assets/assets-and-images

So in pubspec under assets you should add:
- assets/images/

fres-sudo: Ur right, I'll fix it

lib/pages/home_page.dart
```diff
@@ -242,32 +242,33 @@ class _HomePageState extends ConsumerState<HomePage> with Functions {
...
                       physics: const BouncingScrollPhysics(),
                       scrollDirection: Axis.horizontal,
                       separatorBuilder: (context, i) =>
-                          const SizedBox(width: 16),
+                          const SizedBox(width: Sizes.lg),
                       itemBuilder: (context, i) {
                         if (i == accounts.length) {
                           return Container(
-                            constraints: const BoxConstraints(maxWidth: 140),
+                            constraints: const BoxConstraints(maxWidth: Sizes.xs),
```
lucaantonelli, line 255: This actually makes the button so small is not even visible, we should remove constraints or set fixed width as before?

fres-sudo: No it's actually a mistake, I'll fix it.

lib/utils/app_theme.dart
```diff
@@ -246,7 +247,7 @@ class AppTheme {
     ),
     tabBarTheme: TabBarTheme(
       indicator: BoxDecoration(
-        borderRadius: BorderRadius.circular(50),
```
lucaantonelli, line 249: Putting shape circle make it looks pretty bad, should be reverted maybe using Sizes.borderRadius * 10 instead of 50 (also for dark mode below)
circle_shape

lib/pages/transactions_page/widgets/month_selector.dart
```diff
@@ -58,7 +59,7 @@ class MonthSelector extends ConsumerWidget with Functions {
         height: currentHeight,
         decoration: BoxDecoration(
           color: Theme.of(context).colorScheme.tertiary,
-          borderRadius: BorderRadius.all(Radius.circular(10)),
+          borderRadius: BorderRadius.circular(Sizes.borderRadius * 10)
```
lucaantonelli, line 62: This should be just Sizes.borderRadiusLarge you can see in the image in my other comment, the figma doesn't have full rounded border in this widget.

lib/utils/app_theme.dart
lucaantonelli: @fres-sudo oh sorry i meant it also need to be done for light theme (row 38).

fres-sudo: No actually it's my fault ahahha, my bad

lib/ui/widgets/accounts_sum.dart
```diff
@@ -48,15 +49,15 @@ class AccountsSum extends ConsumerWidget with Functions {
...
               child: Row(
                 mainAxisAlignment: MainAxisAlignment.spaceBetween,
                 spacing: 8,
                 children: [
                   RoundedIcon(
                     icon: accountIconList[account.symbol],
                     backgroundColor: accountColorListTheme[account.color],
-                    padding: const EdgeInsets.all(6.0),
+                    padding: const EdgeInsets.all(Sizes.xs),
                     size: 20,
```
lucaantonelli, line 61: I think here something like that looks better:
padding: const EdgeInsets.all(Sizes.sm),
size: Sizes.xl,

lucaantonelli (approved):

docs/index.md
```diff
@@ -16,7 +16,7 @@ nav_order: 1
 </div>
 
 <div>
-    <img src="./assets/sossoldi project cover.jpg" width="1000" alt="Sossoldi icon">
```
lucaantonelli, line 19: One last thing, i searched a bit and i don't think the docs (github pages) can access the root folder, so parent assets folder. maybe is better to mantain the images inside the asstets folder of the docs directory.

fres-sudo: Sorry for late reply. You're right, I fixed it!

Issue #352, opened 2025-03-20
Charts enhancements


matteoberla:
I use a lot charts on this type of applications and I've some thoughts about how the data is showed.

Graphs page:
-Categories are not sorted by total expense value;
-Year expenses chart (on the bottom) has not a Y axis values so it's basically useless and not that clear
-On the same chart, if the average has an high value overlaps the Year text.

This is my suggestion:
![Image](https://github.com/user-attachments/assets/01a7d611-82f5-4847-b66d-4fb2aa3eacef)

Account details page:
-there's no place where we can see the yearly trend chart of the account, in this page I can only see the monthly trend.

This is my suggestion:
![Image](https://github.com/user-attachments/assets/74adf3c8-1403-4646-9910-5a09facb7305)

If you want to assign me this issue I'll be happy to make these changes!

theperu:
@matteoberla thank you, I assigned the issue to you but I would wait for a feedback from the designers before implementing the changes. We'll comment here with the suggestions once we have them

matteoberla:
No problem, I'll wait for your feedback before proceeding with the development

matteoberla:
Hi guys, any news from the designer team on this?

federicopozzato:
@matteoberla Free to proceed! We have in the backlog the enrichment of the data visualization, but in the meantime you can proceed.
Also, if it helps you, there is already in the figma a recap page for the account with a chart https://www.figma.com/design/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?node-id=3014-7173&t=H2yXEO7XP747pMlx-11

PR #387, opened 2025-03-27
Chart enhancement as discussed in issue #352

matteoberla:
## 🎯 Description

Closes: #352   

## 📱 Changes

-Edited the account main chart to show year monthly balance;
-Added currency char to total account amount;
-Categories ordered by amount DESC;
-Added Year expense Y axis title;

## 🔍 Checklist for reviewers
- [ ] Tests are passing
- Tested on:
    - [ ] iOS
    - [ ] Android

### DIFF
--- a/lib/model/bank_account.dart
+++ b/lib/model/bank_account.dart
@@ -336,4 +336,55 @@ class BankAccountMethods extends SossoldiDatabase {
 
     return result;
   }
+
+  Future<List> accountMonthlyBalance(
+      int accountId, {
+        DateTime? dateRangeStart,
+        DateTime? dateRangeEnd,
+      }) async {
+    final db = await database;
+
+    final accountFilter =
+        "(${TransactionFields.idBankAccount} = $accountId OR ${TransactionFields.idBankAccountTransfer} = $accountId)";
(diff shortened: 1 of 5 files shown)


napitek:
Thanks for the contribution!
The categories sorted so much better to see than before, I just don't understand the use of `SplayTreeMap` (seems a bit overkill to me). It works, but since you consume the data in a `ListView`, I'd say a `Map.fromEntries` was more than fine.
There is a specific reason that escapes me?
The rest seems okay to me.

if you can solve the couple of conflicts, let's do a test and then merge it.

matteoberla:
@napitek 
My bad, I'm used to use that method but in this case it's not that convenient.
I've also noticed that when two or more categories have the same amount the `SplayTreeMap` collapses all in one category. I've changed the method and added a sorting by name.

napitek:
@matteoberla Hi, could you check the conflicts when you have time so we merge it?

matteoberla:
I've made all the changes almost a month ago, I thought it was all good, but I'll re check for all the conflicts tomorrow!

matteoberla:
Should be fixed now!

napitek (approved):

PR #407, opened 2025-04-02
Make app version shown in the UI dynamic

mikev-cw:
## 🎯 Description

With these edits, the App Version shown in the UI is taken from the value written in the pubspec file, instead of hardcoding it.

## 📱 Changes

- Bumped app version to 1.0.6
- Added package_info_plus dependency
- edited more_info.dart to pick the right version

## 🧪 Testing Instructions
Just check the app Info page!

## 🔍 Checklist for reviewers
- [ ] Tests are passing
- Tested on:
    - [ ] iOS
    - [ ] Android


pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: A wealth management / personal finance / net worth tracking app.
 
 publish_to: 'none'
 
-version: 0.1.2+3
+version: 0.1.6+1
```
lucaantonelli, line 6: the version is ok to be upgraded but also the code version must always go up (this is required by both the stores) we are now at 0.1.5+6 so version now should be 0.1.6+7

fres-sudo:
I would wait for the merge of this PR tbh. This code need to be revisited. We cannot use the package directly inside the UI like that.

We should provide a `service` that handles the communication with the package, eventually a `repository` that communicates between the `provider` and the `service`, and as you can guess also a `provider` that holds the state of the information that we need. 

(optional) About the information that we need, we should also create a model that is our entity inside the app that matches excactly what we need (ex: appVersion, or others) instead of using the object given from the package.

This process is crucial to ensure that if one day `package_info_plus` won't be the standard anymore we just need a drop-in replacement inside the service and our logic and ui won't be touched at all.

So I would wait for the tomorrow meeting in order the establish an architecture (that is potentially different from the one above) and stick with it, otherwise we would then have to refactor also this part.

mikev-cw:
@fres-sudo Yes, that's the right approach i guess. Let's wait for our meeting then 👍🏻

PR #414, opened 2025-04-06
fix,ref(ci/cd): only one workflow and dart format fix

fres-sudo:
## 🎯 Description

Before this PR there were 3 workflows that did similar things, leading to significant repetition.

The workflow for the format check did not use fvm leading to inconsistencies and errors in the PR.

After this PR there is only one workflow for the CI/CD and the apk build is done if and only if it is a push on the main branch, as before the APK build was done on the PR as well.

## 📱 Changes

- Removed `dart_format_check.yml`
- Removed `test.yml`
- Renamed `flutter-ci.yml`to `ci-cd.yml`

## 🔍 Checklist for reviewers
- [x] Tests are passing
- Tested on:
    - [x] iOS
    - [x] Android


.github/workflows/ci-cd.yml
```diff
@@ -0,0 +1,52 @@
+name: Sossoldi Flutter CI
+
+on:
+  schedule:
+    - cron: "0 0 * * 1"
```
dariowskii, line 5: @fres-sudo I think we can remove the cron setup

.github/workflows/ci-cd.yml
```diff
@@ -0,0 +1,52 @@
...
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+
+jobs:
+  build-android-apk:
+    runs-on: ubuntu-latest
+
```
dariowskii, line 14: @fres-sudo We can add
```yaml
jobs:
  build-android-apk:
    runs-on: ubuntu-latest
    strategy:
      fail-fast: true
```

to cancel the workflow on failure

.github/workflows/ci-cd.yml
```diff
@@ -0,0 +1,52 @@
+name: Sossoldi Flutter CI
+
+on:
+  schedule:
+    - cron: "0 0 * * 1"
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
```
dariowskii, line 9: @fres-sudo I would personally remove the push trigger, and add the `develop` branch for future PRs

```yaml
on:
  pull_request:
    branches: [ "master", "develop" ]

concurrency: 
  group: ${{ github.head_ref }}
  cancel-in-progress: true // This cancel the running workflow if there are new commits to the PR
```

.github/workflows/ci-cd.yml
```diff
@@ -0,0 +1,52 @@
...
+          fi
+
+      - name: Run tests
+        run: flutter test --concurrency=1
+
+      - name: Build APK and App Bundle
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        run: |
+          flutter build apk
+          flutter build appbundle
```
dariowskii, line 52: @fres-sudo I think we can keep only the appbundle build, just because is the best choice for Play Store

dariowskii (changes requested):

dariowskii:
@fres-sudo For now is ok, but these little changes will make a smarter CI. Also, we can optimize more with a matrix strategy and concurrent steps but we can do it later

fres-sudo:
@dariowskii Thanks for the review!

I made the changes 👍

.github/workflows/ci-cd.yml
```diff
@@ -0,0 +1,55 @@
...
+            exit 1
+          else
+            echo "✅ Code format check passed."
+          fi
+
+      - name: Run tests
+        run: flutter test --concurrency=1
+
+      - name: Build APK and App Bundle
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```
dariowskii, line 53: @fres-sudo we want to keep the build check only for the push on main? For me, this is an essential step because any PR can destroy the build (also if tests passes)

fres-sudo: I thought it might ease the PR ci. I see that it can cause errors, even if test passes and so on, but the chances are very very low, and if you ever push on main and something breaks we would still have time to “fix it,” in the sense that on main you don't go directly into production like when u work on web, the release anyway, for now, we do it by hand.
Idk, what do u think @lucaantonelli

.github/workflows/ci-cd.yml
VoidElle: I'm not sure if also external stores like F-Droid (Or others that community members requested) also prefer .appbundle instead of .apk

.github/workflows/ci-cd.yml
fres-sudo: > how do you reconstruct the path of what broke once so many PRs are pushed from dev to main

Having the `dev` branch is not something necessary at all. We can easily avoid it, we don't need to push a bunch of commits from `dev` to `main` for a new release this is even more true from the time that we don't have any CD for the stores publishing, so having two branches is completely useless now.

We can "define" a new release from whenever we push to the stores directly.

And for my exp this model works very, very well, and it scales a lot: so having `main` (or `master`) branch to push directly and then just works with PRs that targets only `main`.

And yes, it will come handy for the big refactoring we'll do in which we actually need a parallel branch to avoid headaches, but still it's a niece case...and if we don't touch any `ios` or `android` related files, or we don't add critical packages (that modify some native codes) it's very unlucky to have issues.

Plus, usually when the build fails, the errors are very precise.

If you prefer, we can include the build when we **push** on `main`, but IMHO it's avoidable for PRs.

dariowskii: I disagree with this...a second development branch in my opinion is useful for several reasons:

- it allows you to check for errors in advance before they are put on the main branch
- from a development branch you can pick any commits to do cherry-picks
- you can set up pipelines to trigger releases only if a tag is made on the main branch (for example, with Xcode Cloud)
- it allows you to have a clear distinction between what is in development and what is stable and released

(and so on...)

It's probably schools of thought here too, but I really think that in order not to go crazy in the future you need to have at least two branches.

Let's hear other opinions as well!

VoidElle: @fres-sudo I agree with some of @dariowskii's topics.

With a development branch we could have master always aligned with what is currently released in production, and I think with this approach we might have a better disaster recovery. 

We could have a broken development branch, but a safe and working master branch that is always aligned with production (and push there only on new, and heavy tested code, for releases)

.github/workflows/ci-cd.yml
```diff
@@ -0,0 +1,55 @@
...
+            exit 1
+          else
+            echo "✅ Code format check passed."
+          fi
+
+      - name: Run tests
+        run: flutter test --concurrency=1
+
+      - name: Build APK and App Bundle
+        if: github.event_name == 'push'
```
dariowskii, line 53: This `if` doesn't work for PR

dariowskii (changes requested):

.github/workflows/ci-cd.yml
fres-sudo: wdym?
If u mean that the build is not triggered on PRs, this was the intended behavior.

As discussed yesterday, since we have a dev branch, where we can recover stuffs, do heavy tests and stuffs, before pushing into main branch, I don't think we then need the build on PRs

.github/workflows/ci-cd.yml
mikev-cw: WDYT if we add this dev branch topic on the next meeting agenda?

dariowskii (approved):

PR #415, opened 2025-04-06
UI changes regarding issue #400

Tornado-Hacker:
## 🎯 Description

The Total UI of the older_recurring_payment was changed.

Closes: # 400

## 📱 Changes

- [ ] Changed the whole  UI 
- [ ] For Monthly intervals it shows --> "Monthly - on the 4th" but for remaining like days, quarterly it shows in days --> "Daily in  day"
- [ ] In the design part there was grey background for older payments heading it does not fit properly so I removed that.
- [ ] The commit shows I had made changes to build gradle but there was no issue it was the same file from the main repo.
- [ ] For height of modalbottomsheet i added height in recurring_payment_cart.dart file.

## 📸 Screenshots / Screen Recordings (if applicable)

changes_at_intervals
with_grey
without_grey

Issue #400, opened 2025-03-28
Recurrent transactions list page not designed correctly


federicopozzato:
The recurring transactions list page (that appear when I go in Planning > recurring transactions section > see older payments) is not designed as in Figma.

**How it is:**
![Image](https://github.com/user-attachments/assets/970f3779-05bd-4c1e-957a-623c0c7b3898)

**How it should be:**
https://www.figma.com/design/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?node-id=1909-4728&t=V1f3Hcl4SS33x9fA-11
![Image](https://github.com/user-attachments/assets/c1ada5d0-cdcd-41e8-b675-20fcd9829864)

Tornado-Hacker:
Hey, I would like to work on this widget.

PR #430, opened 2025-04-15
UI changes regarding issue #400 #415

Tornado-Hacker:
## 🎯 Description

This is the new pr request with needed changes from pr request #415


lucaantonelli:
Hi, thank for the PR.
I think it needs some modifications, first you should make a rebase to get latest code update.
Then, the modal is not dismissible, neither the back button is working (you should use a textbutton instead of a button only on the icon).

theperu:
Hi @Tornado-Hacker👋

Thanks!

Issue #468, opened 2025-09-20
[Bug] Support old backups that have no type for recurringTransaction


theperu:
### 🛰️ **What happened?**

After enabling recurring transactions for income and transfers the backups created before can't be imported because they are missing the type (like IN, OUT, TRSF)

### 🖥️ **OS**

Both

### 🔄 **What version?**

All

### 🔢 **Version**

Most recent on main branch

### 📝 **Code of Conduct**

- [x] ✅ I agree to the Code of Conduct

PR #471, opened 2025-09-30
fix (migration): type in recurring real to text

lucaantonelli:
## 🎯 Description

After enabling recurring transactions for income and transfers the backups created before can't be imported because they are missing the type (like IN, OUT, TRSF)

Closes: #468 

## 📱 Changes

- Add a migration that fix the type of column type from real to text and set default to out, so also during import it does get it by default if the field is empty.

### Behaviour
1. Create a backup with the current version of the app in the stores
2. Install this version of the app
3. Try to import the backup and make sure that the action completes without errors

## 🔍 Checklist for reviewers
- [x] Tests are passing
- Tested on:
    - [ ] iOS
    - [x] Android

### DIFF
--- a/lib/model/recurring_transaction.dart
+++ b/lib/model/recurring_transaction.dart
@@ -26,6 +26,7 @@ class RecurringTransactionFields extends BaseEntityFields {
     toDate,
     amount,
     note,
+    type,
     recurrency,
     idCategory,
     idBankAccount,
--- a/lib/services/database/migrations/0003_recurring_transaction_type.dart
+++ b/lib/services/database/migrations/0003_recurring_transaction_type.dart
@@ -11,11 +11,11 @@ class RecurringTransactionType extends Migration {
 
   @override
   Future<void> up(Database db) async {
-    const realNotNull = 'REAL NOT NULL';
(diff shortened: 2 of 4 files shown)


theperu (changes requested):
@lucaantonelli I checked this but there seems to be an issue with the migration. 
### Steps to reproduce
1. I tried to install an old version of the app (commit b1f1f0ca393968b12a6757c84628defcab11819a) 
2. Filled the DB with the demo data from settings
3. Installed the app with your changes
4. When I open I see this error (both on iOS and Android):
Screenshot_2025-09-30-13-19-56-547_com ripster sossoldi

lucaantonelli:
> @lucaantonelli I checked this but there seems to be an issue with the migration.
> 
> ### Steps to reproduce
> 1. I tried to install an old version of the app (commit [b1f1f0c](https://github.com/RIP-Comm/sossoldi/commit/b1f1f0ca393968b12a6757c84628defcab11819a))
> 2. Filled the DB with the demo data from settings
> 3. Installed the app with your changes
> 4. When I open I see this error (both on iOS and Android):
>    Screenshot_2025-09-30-13-19-56-547_com ripster sossoldi

Now it should work fine

PR #483, opened 2025-11-25
add accessibilityId for create edit account page

fedemic:
## 🎯 Description

Added Semantics and unique labels to key widgets, providing `accessibilityId` on iOS and `contentDescription` on Android.
This improves element identification for Appium and significantly simplifies the development of automated E2E tests.

## 🔍 Checklist for reviewers
- [x] Tests are passing
- Tested on:
    - [ ] iOS
    - [x] Android

### DIFF
--- a/lib/pages/accounts/create_edit_account_page.dart
+++ b/lib/pages/accounts/create_edit_account_page.dart
@@ -56,8 +56,11 @@ class _CreateEditAccountPage extends ConsumerState<CreateEditAccountPage> {
     return Scaffold(
       appBar: AppBar(
         title: Semantics(
-          label: "create_edit_account_title",
-          child: Text("${selectedAccount == null ? "New" : "Edit"} account"),
+          label: "create_edit_account",
+          container: true,
+          child: ExcludeSemantics(
+            child: Text("${selectedAccount == null ? "New" : "Edit"} account"),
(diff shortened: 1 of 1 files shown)


lucaantonelli:
Hi, I think it could be a nice addition, but I'm not sure about using Semantics at this point in the project. I think it could be better to add it at the end of the beta phase. For e2e testing, the key-value properties of widgets should be enough, without needing to wrap every required widget with Semantics. The problem, at least in my mind, is that since a lot is still in progress and may change, wrapping widgets in Semantics every time could add complexity and worsen code readability. But let’s hear what the others think and decide together.

fedemic:
Hi @fres-sudo and @lucaantonelli, thanks for reviewing and sharing your thoughts.

I initially went with `Semantics` because it’s a common pattern for adding accessibility identifiers in Flutter. However, I agree that it would make the code heavier and negatively affect maintainability.
My only concern with using key-value is that it's not officially supported by our test automation framework, but I think it’s acceptable to start developing scripts based on this approach for now, keeping in mind that a later refactor (e.g. wrapping widgets for proper accessibility support) would benefit both accessibility and testing.

I’ll update the implementation as suggested in this PR, move it back to draft, and request a new review once the changes are ready.

Issue #479, opened 2025-11-20
[Feature] Ability to reorder the category list


phidotexe:
### 📝 **Feature Description**

Adding a property `order` to the `category_transaction` and handling the drag & drop functionality in order to customize the list of categories

### 🛑 **Problem Statement**

I have been using the app for a whie and I've decided to make some changes to the categories I want to track. 
I want to be able to have the list in a set order so that I can find elements more easily and "group them" together. 

### 🎁 **Benefits**

No need to delete categories and recreate them in a defined order

### 🖥️ **OS**

Both

### 📜 **Code of Conduct**

- [x] ✅ I agree to the Code of Conduct

PR #486, opened 2025-12-10
feat: Add drag & drop reordering for  categories

Kartik30R:
## 🎯 Description

This PR enables users to customize the order of transaction categories via drag & drop.

* **UI:** Implemented `ReorderableListView` in the category settings.
* **Database:** Added a `position` column via Migration 0004 to persist the order.
* **Logic:** Field is named `position` (instead of `order`) to avoid SQL keyword conflicts.

Closes: #479

## 📱 Changes

* Created database migration `0004_add_category_position.dart` to add the `position` column with safe defaults for existing users.
* Refactored `CategoryList` widget to replace `ListView` with `ReorderableListView.builder`.
* Updated `CategoryTransaction` model to include the `position` field.
* Implemented `reorderCategories` in `AsyncCategoriesNotifier` (Provider) to handle optimistic UI updates and background database syncing.

### Behaviour
1. Navigate to **Settings -> Categories**.
2. Long press on any category card.
3. Drag and drop it to a new position in the list.
4. Kill the app completely (remove from recent apps).
5. Relaunch the app and verify that the category order is preserved.

 https://github.com/user-attachments/assets/ae627fa2-289b-48d5-8265-1ab02dd031e5

## 🔍 Checklist for reviewers
- [ ] Tests are passing
- Tested on:
    - [ ] iOS
    - [x] Android

## ✍️ Additional Context

I used the column name `position` instead of `order` because `ORDER` is a reserved keyword in SQLite, which causes syntax errors in queries unless strictly escaped. Using `position` is safer and cleaner for future maintenance.


theperu:
@Kartik30R thanks for the contribution! I'll have a look at the pr tomorrow 😁

lib/model/category_transaction.dart
```diff
@@ -6,6 +6,7 @@ const String categoryTransactionTable = 'categoryTransaction';
 
 class CategoryTransactionFields extends BaseEntityFields {
   static String id = BaseEntityFields.getId;
+  static String order = 'position';
```
lucaantonelli: You could call the field name order as well instead of position, for more consistency.

lucaantonelli (changes requested):
Hi, thanks for the PR, is really good. I left some points to review, other than that, do you think could be made the same as well for the accounts section? Should be almost the same, this is to just have a single db migration. Also, when you have finished run the command "dart format ." before pushing other commits, since this one is not formatted. If you have any questions feel free to ask here or on our discord.

Kartik30R:
> Hi, thanks for the PR, is really good. I left some points to review, other than that, do you think could be made the same as well for the accounts section? Should be almost the same, this is to just have a single db migration. Also, when you have finished run the command "dart format ." before pushing other commits, since this one is not formatted. If you have any questions feel free to ask here or on our discord.

@lucaantonelli I’ll complete the changes for this PR first . I’ll handle the account section updates later in a separate PR mainly to keep this one scoped and manageable

lib/model/category_transaction.dart
lucaantonelli: You are right, good point. Then yes, leave it as position in the db, will be fine anyway.

Issue #353, opened 2025-03-20
UI Bug on the transactions lists


RiccardoCherchi:
## Description

I founded e little UI bug on the transactions lists of the app, if the name of the transaction is too long it will overflow from the card. Wrapping the text inside an Expanded widget should do the work

## Screenshots

![Image](https://github.com/user-attachments/assets/c900a474-b913-4827-9812-8a77ccf250bd)

![Image](https://github.com/user-attachments/assets/bca6275b-1e02-4c66-ad46-a0daed8dee6f)

theperu:
@RiccardoCherchi I assigned it to you since it's a simple thing you can either work on it and then show the solution on a PR to the Design team or wait for their feedback first

RiccardoCherchi:
After installing the development app on my emulator i notices the bug has already been fixed

![Image](https://github.com/user-attachments/assets/32113b25-5455-408e-ab3f-3af6024046a3)

But there is another UI overflow on the Transactions > Categories/Accounts page, i will send a PR for that, i think doing this in this already opened issue is better

PR #487, opened 2025-12-14
Fix/minor UI fixes

theperu:
## 🎯 Description

- Income categories in the graphs page were sorted from lowest to highest amount, making it harder to see the most significant at a glance. Now sorted from highest to lowest.
- Search results were not scrollable, preventing users from viewing all matching transactions.
- Long transaction descriptions caused text overflow in the categories/accounts tab panels.
- Currency selector dialog had inconsistent background colors between the dialog and list items.

**NOTE:** For the currency selector it was proposed a different approach but since we are working on a redesign I think it's just easier to fix the issue and then we will rework that part in the future if we decide it's needed

Closes: #356 #353 

## 📱 Changes

- Sort categories by absolute value in descending order (highest to lowest) in graphs page
- Wrap search results in `SingleChildScrollView` to enable scrolling
- Add overflow: `TextOverflow.ellipsis` and date placeholder for empty notes in `panel_list_tile.dart`.
- Set `tileColor: Colors.transparent` on ListTile in currency selector dialog

### Behaviour
1. Navigate to Graphs page and verify categories are sorted from highest to lowest amount
2. Go to Search page, perform a search with many results, and verify you can scroll through all results
3. Navigate to Transactions page, switch to Categories or Accounts tab, expand a category/account with transactions that have long notes or no notes, verify no text overflow and empty notes show date instead
4. Go to Settings -> General Settings, tap on currency selector and verify the list items have a consistent background with the dialog

## 🔍 Checklist for reviewers
- [ ] Tests are passing
- Tested on:
    - [ ] iOS
    - [ ] Android

### DIFF
--- a/lib/pages/transactions/widgets/panel_list_tile.dart
+++ b/lib/pages/transactions/widgets/panel_list_tile.dart
@@ -134,8 +134,9 @@ class PanelListTile extends ConsumerWidget {
                                   Expanded(
                                     child: Text(
                                       (transaction.note?.isEmpty ?? true)
-                                          ? DateFormat("dd MMMM - HH:mm")
-                                              .format(transaction.date)
+                                          ? DateFormat(
+                                              "dd MMMM - HH:mm",
+                                            ).format(transaction.date)
                                           : transaction.note!,
                                       overflow: TextOverflow.ellipsis,
                                       style: Theme.of(


lucaantonelli:
LGTM! :)

Issue #247, opened 2025-03-13
Empty states


federicopozzato:
When I first used te app I noted that the empty states of the main pages were not correctly set with the illustrations, copy and CTA showed in Figma.
Please check

Here some references:
https://www.figma.com/design/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?node-id=3242-8937&t=wIcBgzLJtYVqsUqS-11
https://www.figma.com/design/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?node-id=3020-7707&t=wIcBgzLJtYVqsUqS-11
https://www.figma.com/design/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?node-id=976-2601&t=wIcBgzLJtYVqsUqS-11
https://www.figma.com/design/6NyY9yqunpbU7HIkbNEAL3/Sossoldi-App?node-id=2877-5431&t=wIcBgzLJtYVqsUqS-11

If some empty states are missing in Figma, please let me know!

uveshm003:
Home Screen & Graph Screens empty states are remaining.

federicopozzato:
> Home Screen & Graph Screens empty states are remaining.

What do you mean?

uveshm003:
There should be an empty state when graph data empty but instead we are displaying only text as of now.

uveshm003:
@theperu  Added the PR for this issue.

uveshm003:
Can you please assign this to me.

theperu:
@uveshm003 take a look at the figma and follow the approaches that are there

uveshm003:
@theperu  already completed. You can check the #300 PR.

theperu:
Did you implement all states suggested here? If so we'll try to review the PR as soon as possible, we are getting a lot of contributions these days so it might take a bit 😅

uveshm003:
I have added for transaction list & The budget screen was already added. I only found these 2 in figma.

federicopozzato:
Sorry, I don’t get what you’re saying. As I see, the empty states you’re talking about are present in figma. If something is missing please let me know, just be clear please. Thanks!

PR #493, opened 2026-01-10
fix: transactions empty state

lucaantonelli:
## 🎯 Description

Add the transactions empty state 

Closes: #247 

## 🔍 Checklist for reviewers
- [x] Tests are passing
- Tested on:
    - [ ] iOS
    - [x] Android

### DIFF
--- a/lib/pages/planning/widget/budget_card.dart
+++ b/lib/pages/planning/widget/budget_card.dart
@@ -143,17 +143,23 @@ class BudgetCard extends ConsumerWidget {
                       child: TextButton.icon(
                         icon: Icon(
                           Icons.add_circle,
-                          color: Theme.of(context).colorScheme.secondary,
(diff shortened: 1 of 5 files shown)


theperu (changes requested):
Currently I can see that the buttons for the empty states are hardcoded to be white both on the dark and light theme but I'm not sure this is the expected outcome. @federicopozzato WDYT?
Reference:
Screenshot 2026-01-12 at 20 28 10
Screenshot 2026-01-12 at 20 28 16

federicopozzato:
@theperu I think because otherwise the cta color is not very different from the background. But it's the same in light mode, so I suggest to follow the current light/dark combination. This will for sure be fixed in the future.

lucaantonelli:
@theperu I fixed it!

theperu (approved):
LGTM!

Issue #198, opened 2025-01-28
Feature: Subcategories


theperu:
# 🙋🏻‍♂️ Add subcategories

## 📋 Description

**Summary**:  
We should allow a user to track their expenses also with subcategories. Ex. Category: groceries, Subcategory: Coop

**Expected Behavior**:  
From the settings or the new transaction page the user should be able to add subcategories to an existing category, and he should be able to use it in the new transactions.

**Notes**:  
Lets consider how to handle them both in the frontend (what data do we want to show about these?) and DB (how do we want to store them?)

PR #494, opened 2026-01-12
feat: subcategory feature

lucaantonelli:
## 🎯 Description

Working implementation of subcategory feature!

Closes: #198

## 📱 Changes

- [x] Subcategories are shown as a list on each category page (following figma)
- [x] Subcategories creation/editing page (only allow to choose name and icon, not color, since is the same of the parent category)
- [x] Subcategories can be selected in transactions. Click on a category with subcategories, and it will shows with an animated list the subcategory, if you don't choose a subcategory, the parent is automatically selected
- [x] Subcategories doesn't appear in the main category list, but only inside the parent editing page

## 🧪 Testing Instructions

Test the pages mentioned in the changes.

## 🔍 Checklist for reviewers
- [x] Tests are passing
- [] New tests are added (if needed)
- Tested on:
    - [ ] iOS
    - [x] Android

### DIFF
--- a/lib/pages/transactions/widgets/categories_tab.dart
+++ b/lib/pages/transactions/widgets/categories_tab.dart
@@ -17,67 +17,8 @@ class CategoriesTab extends ConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    final categories = ref.watch(categoriesProvider);
-    final transactions = ref.watch(transactionsProvider);
+    final categoriesData = ref.watch(categoryWithSubcategoriesDataProvider);
     final transactionType = ref.watch(selectedTransactionTypeProvider);
-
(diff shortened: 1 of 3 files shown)


theperu (changes requested):
@lucaantonelli I noticed a few things while testing:
- If I try to create multiple subcategories the page for creating them is already prefilled with the values from the previous one. (https://github.com/user-attachments/assets/33d5ce35-ba5b-407b-a2be-6deed3614923)
- In the transactions page I get an error after adding a transaction with a subcategory
Screenshot 2026-01-18 at 20 11 03

- If I add a recurring transaction with a subcategory then it doesn't show in the planning page (https://github.com/user-attachments/assets/25b4aa48-a138-4a1b-a400-c2a21fdf9960)

PR #496, opened 2026-01-13
Persist last used value for balance blurring in SharedPreferences

mikev-cw:
## 🎯 Description

Sorry for closing [the other PR](https://github.com/RIP-Comm/sossoldi/pull/490), i messed up with my git repo after merging #489 

Blurring balance was always active by default, every time you open the application. Let's save the last used value on sharedpref, and use that instead.

## 📱 Changes

- Passed the sharePreferences dependency on main to the visibilityAmountProvider
- Make the settings read from that

### Behaviour
1. Tap balance visibility button to see/blur values
2. Close the app
3. Open the app, check if the last value is used

## 🔍 Checklist for reviewers
- [x] Tests are passing
- Tested on:
    - [x] iOS
    - [x] Android

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -88,9 +88,6 @@ void main() async {
           overrides: [
             versionProvider.overrideWithValue(packageInfo.version),
             sharedPrefProvider.overrideWithValue(sharedPreferences),
-            visibilityAmountProvider.overrideWith(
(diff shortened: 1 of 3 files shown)


lucaantonelli:
LGTM!

PR #520, opened 2026-03-05
Implemented import from MoneyManager

Mattia-Sacchi:
I have added the ability to import a CSV file from the MoneyManager application to the app:
https://play.google.com/store/apps/details?id=com.realbyteapps.moneymanagerfree&hl=en-US&pli=1

I have been using this application for two years and wanted to migrate to yours, so I added this feature. I have already tested the import with data from both my own account and those from my girlfriend and other friends who use the app.

Among the main changes, there is the addition of a button in the backup settings and the implementation in the database, which is systematically overwritten. However, there would be a possibility to change the code a bit to integrate the input data from a certain date to another.

It adds all categories and subcategories as well as the accounts that come from MoneyManager, providing an excellent migration experience.
If there are any issues with this feature, I am willing to monitor the app's issues and resolve them myself.

P.S. I have added myself to the collaborators' screen in case you accept the PR. 

immagine


Mattia-Sacchi:
I’ve also added two buttons to import/export the raw database, which are significantly faster.

theperu (changes requested):
Nice work! I just have a few notes here and there:
- There are some conflicts that need to be resolved before merging
- I noticed that the exported csv has different names for the columns based on the language and at the moment it just works if you use the English one
- The icons are all the same, ideally we should have different ones for each option, ex. for the MoneyManager import we could have a pig, for the db options we could use a DB icon (if we can even find different ones to use for the import and export that would be ideal)

PR #524, opened 2026-03-25
Feat: Category and Account deletion

### LINKED ISSUE #177: Bug when deleting a category

lucaantonelli:
## 🎯 Description

Handle categories and accounts deletion correctly. Add a deleted_at field on both tables.

Closes: #177 

## 📱 Changes

- [x] Add migration for deleted_at fields in categories and bank accounts tables
- [x] Implement soft delete, to avoid breaking transactions.

## 🧪 Testing Instructions

- Test deleting both categories and accounts, check if everything is still working.

## 🔍 Checklist for reviewers
- [x] Tests are passing
- Tested on:
    - [ ] iOS
    - [x] Android

### DIFF
--- a/lib/model/bank_account.dart
+++ b/lib/model/bank_account.dart
@@ -1,3 +1,4 @@
+import '../ui/extensions.dart';
 import 'base_entity.dart';
 
 const String bankAccountTable = 'bankAccount';
@@ -15,6 +16,7 @@ class BankAccountFields extends BaseEntityFields {
   static String order = 'position';
   static String createdAt = BaseEntityFields.getCreatedAt;
   static String updatedAt = BaseEntityFields.getUpdatedAt;
+  static String deletedAt = BaseEntityFields.getDeletedAt;
 
   static final List<String> allFields = [
     BaseEntityFields.id,
@@ -28,6 +30,7 @@ class BankAccountFields extends BaseEntityFields {
     order,
     BaseEntityFields.createdAt,
     BaseEntityFields.updatedAt,
+    BaseEntityFields.deletedAt,
   ];
 }
 
@@ -55,6 +58,7 @@ class BankAccount extends BaseEntity {
     this.total,
     super.createdAt,
     super.updatedAt,
+    super.deletedAt,
   });
 
   BankAccount copy({
@@ -69,6 +73,7 @@ class BankAccount extends BaseEntity {
     int? order,
     DateTime? createdAt,
     DateTime? updatedAt,
+    DateTime? deletedAt,
   }) => BankAccount(
     id: id ?? this.id,
     name: name ?? this.name,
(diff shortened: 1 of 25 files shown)


theperu (changes requested):
Hi! There are a few things that I noticed:
1. The migration  version is 6 instead of 7
2. When I delete a category there is no dialog similar to the one that we have for the accounts
3. When I delete a category that has a recurring transaction associated the category is deleted but the recurring transaction is not
4.  When I delete an account that has a recurring transaction associated the account is deleted but the recurring transaction widget does not update until I close and reopen the app
5. When I delete an account that is set as main account it gets deleted but I still see it as the default option when I add a new transaction (this probably happens also to the counts towards net worth BTW)
6. When I delete a category/subcategory I still see it in the list of possible categories when I add a new transaction, the list updates after closing and reopening the app
7. If I delete a category that has a budget when I go back to the planning page I see an error

lucaantonelli:
> Hi! There are a few things that I noticed:
> 
> 1. The migration  version is 6 instead of 7
> 2. When I delete a category there is no dialog similar to the one that we have for the accounts
> 3. When I delete a category that has a recurring transaction associated the category is deleted but the recurring transaction is not
> 4. When I delete an account that has a recurring transaction associated the account is deleted but the recurring transaction widget does not update until I close and reopen the app
> 5. When I delete an account that is set as main account it gets deleted but I still see it as the default option when I add a new transaction (this probably happens also to the counts towards net worth BTW)
> 6. When I delete a category/subcategory I still see it in the list of possible categories when I add a new transaction, the list updates after closing and reopening the app
> 7. If I delete a category that has a budget when I go back to the planning page I see an error

Hi, all the problem should be fixed, i couldn't replicate the problem at point 5, on my end is working correctly. I also found an error in budgets, if you create a budget and delete it then when recreating it for same category doesn't show anymore because active still remain false when updating, i should have fixed this as well.

theperu (changes requested):
@lucaantonelli I still see a couple of issues:
- I can't delete subcategories
- If I open the new transaction page, open the list of categories, then I go to settings and delete a category when I come back to add a transaction again I still see the deleted category in the list
- When an account is deleted I would set the countNetWorth to false to avoid future issues with it

mikev-cw (changes requested):
Huge work Luca!!! That was not so simple as it seems...

Some things i noticed:
1. account / category deletion permanently deletes recurring transactions even though the UI says they are only “stopped”. We should decide what's better: correct the alert to the user, or actually disable recurring transactions
2. after deleting the main account, new transaction flows can still pick a deleted account as the default
3. Cannot delete subcategories (or they're deleted but still visible somehow, not sure)
4. Removing a category also hard-deletes budgets for that category and its direct subcategories, but the dialog only warns about recurring transactions

theperu (approved):
LGTM! ✅

I checked both the issues reported by me and @mikev-cw and they are solved now! 🎉

PR #532, opened 2026-07-03
Feat: Apple Shortcuts integration

mikev-cw:
## 🎯 Description
This PR adds iOS Shortcuts support for quickly creating wallet transactions through App Intents.
Currently, creating a transaction requires opening the app and manually navigating through the transaction flow. This creates friction, especially for quick everyday expenses.
With this change, users can set up an automation that adds an uncategorized transaction to Sossoldi whenever they pay with their phone using NFC. The transaction can then be reviewed and categorized later inside the app.

I think this could be a low-priority feature, but I'd like to merge it (and maybe silently release) to facilitate testing on a physical device through TestFlight.

## 📱 Changes
- Added an iOS `AddWalletTransactionIntent` (**Warning**: This is Swift code, which isn't really my strong point...)
- Connected the shortcut entry point to the existing transaction flow.
- Updated transaction provider handling so shortcut-created transactions can be saved consistently with in-app transactions even without category.

### Changes applicable only on iOS
Unfortunately, this feature isn't easily testable on a simulator. You need to install the app on a physical device through Xcode. Then: 
1. Go **to Shortcuts** -> **Automation** -> **Add**
2. Choose "Wallet"
3. Select all cards (or only the ones you want to track in Sossoldi) and all categories -> **Next**
4. Search for Sossoldi and tap on the "Add Wallet Transaction" intent. Configure it like this:

Screenshot 2026-07-03 alle 18 19 41

5. Save and go buy something 🤣 

This is a real mobile payment transaction I made. As you can see, an uncategorized transaction appears in Sossoldi with the merchant name for easy recognition.

https://github.com/user-attachments/assets/5c726e58-fc39-4ee0-a164-016fa96053e0

### DIFF
--- a/ios/Runner.xcodeproj/project.pbxproj
+++ b/ios/Runner.xcodeproj/project.pbxproj
@@ -10,12 +10,13 @@
 		1498D2341E8E89220040F4C2 /* GeneratedPluginRegistrant.m in Sources */ = {isa = PBXBuildFile; fileRef = 1498D2331E8E89220040F4C2 /* GeneratedPluginRegistrant.m */; };
 		331C808B294A63AB00263BE5 /* RunnerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 331C807B294A618700263BE5 /* RunnerTests.swift */; };
 		3B3967161E833CAA004F5970 /* AppFrameworkInfo.plist in Resources */ = {isa = PBXBuildFile; fileRef = 3B3967151E833CAA004F5970 /* AppFrameworkInfo.plist */; };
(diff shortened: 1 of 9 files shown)


mikev-cw:
Updated with:

- The iOS intent no longer writes directly to SQLite or duplicates the Dart transaction schema. Instead, it writes a small pending Wallet transaction payload (`amount`, `merchant`, `card`, `createdAt`) into the app container, and Flutter imports it on app start/resume through the existing transaction repository.

- The Dart importer now owns amount parsing, so locale-specific formats such as `1.500` in EU locales are handled correctly. It also preserves refund/credit signs by importing negative amounts as income, keeps Wallet transactions uncategorized and with no associated bank account, and cleans up the pending file after successful import.

- Finally, deleting a Wallet-imported transaction no longer crashes when no selected bank account exists.

- Added "unassigned" account in transactions accounts tab for transactions not assigned to any account yet. Same for unassigned categories.

Issue #481, opened 2025-11-23
[Feature] - Bank account synch using Enable Banking


theperu:
### 📝 **Feature Description**

After talking with [Enable Banking](https://enablebanking.com) we learned that they have a free plan for personal use that we could use for our app. If you want to start experimenting with it you should create an account on their site and try the production environment to link your bank accounts.
This is in early phase so we need to explore how it works, if you have any doubt feel free to post it and I'll forward it to them.

Useful links:
Another project that uses their apis: https://midday.ai/
Documentation: https://enablebanking.com/docs/api/linked-accounts/
Example of how to use their apis: https://github.com/Tawasta/account_bank_statement_import_tilisy and [tilisy.com](https://tilisy.com/en)

### 🛑 **Problem Statement**

Currently we need to all transactions manually, this will help user keep track of at least some. 

### 🎁 **Benefits**

No missed transactions

### 🖥️ **OS**

Both

### 📜 **Code of Conduct**

- [x] ✅ I agree to the Code of Conduct

marianialessandro:
I would like to take care of the implementation of this feature

john-frandsen:
Worth considering [open-banking.io](https://open-banking.io) as an alternative or complement to Enable Banking — it covers the same EU/UK banks (2,600+) but is much simpler to onboard with. No PSD2/eIDAS certificates, no preauthorisation process, just an API key.

Enable Banking is actually one of the providers behind open-banking.io, so we know their API well. One concrete thing we handle on top: we calculate account balances server-side, so you always get a current balance even when the underlying bank's PSD2 API doesn't expose one. Enable Banking on its own passes through whatever the bank gives (or doesn't), which is a known gap for some banks.

Disclosure: I maintain open-banking.io. Happy to help if there's interest in supporting it — either as the provider or alongside Enable Banking. Pricing starts at ~€3/mo and drops as the user base grows.

theperu:
Hi @john-frandsen I appreciate you reaching out, we considered Enable Banking at the moment because the offer also an option for personal use that would be free for our users if they are willing to setup a developer account with them

john-frandsen:
That's a fair point on the free tier — for a self-hosted app where users bring their own credentials, Enable Banking's developer account route makes sense as the default.

[...]

PR #542, opened 2026-09-14
feat(banking): persist bank connections and manage consent lifecycle

marianialessandro:
## Summary

Adds persistence and consent lifecycle management on top of the banking foundation introduced in #536.

- Persist bank connections, provider identity, and pending authorization state, with migrations and compatibility for existing stored data.
- Handle authorization callbacks, interrupted-flow recovery, reconnect, and revocation through provider-neutral banking contracts.
- Integrate credential settings and lifecycle coordination in the application layer, keeping Enable Banking-specific behavior in its adapter.

This is the second PR in the banking integration series for #481. User-facing setup and connection flows will be introduced in the later UI PR.

## Commit structure

1. Validate consent redirects and map connection failures.
2. Persist bank connections and recovery state.
3. Integrate the provider-neutral consent lifecycle.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -4,6 +4,13 @@ analyzer:
   exclude:
     - "**/*.g.dart"
     - "**/*.freezed.dart"
+    - build/**
+    - android/**
+    - ios/**
+    - web/**
+    - windows/**
+    - macos/**
+    - linux/**
   errors:
     invalid_annotation_target: ignore
 
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -1,13 +1,14 @@
 import 'package:flutter/material.dart';
+import 'package:flutter_phoenix/flutter_phoenix.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
-import 'package:package_info_plus/package_info_plus.dart';
 import 'package:intl/date_symbol_data_local.dart';
 import 'package:local_auth/local_auth.dart';
+import 'package:package_info_plus/package_info_plus.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 import 'package:timezone/data/latest.dart' as tz;
 import 'package:timezone/timezone.dart' as tz;
-import 'package:flutter_phoenix/flutter_phoenix.dart';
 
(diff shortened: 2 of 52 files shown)


mikev-cw:
Hey! Thank you so much for this!!! 
I thing is very good overall, just some questions/comments/consideration:

1: In `BankConsentLifecycleService._rejectStagedConnection`, the `REAUTH_REQUIRED` branch only updates the status and returns.
But `BankConnectionRepository.selectAwaitingImport` selects rows only by `pendingRemoteConnectionId IS NOT NULL`.
That means if a staged import fails into `REAUTH_REQUIRED`, its pending fields remain populated, so it can be selected again on every resume.

Should the `REAUTH_REQUIRED` path also clear `pendingRemoteConnectionId`, `pendingAuthorizationId`, and `pendingValidUntil`, or should `selectAwaitingImport` filter by an awaiting/importable status?

2: `BankingBackupPolicy.sanitizeForExport` converts every exported `bankConnection` status to `REAUTH_REQUIRED`.
That seems safe for active sessions because usable session ids are stripped. But for terminal states like `REVOKED` or `DISCONNECTED`, this may make intentionally removed connections appear reconnectable after restore.

Should terminal statuses be preserved while only active/pending session states are downgraded to `REAUTH_REQUIRED`?

3: Consider folding `0009_add_bank_provider_identity` into `0008_add_bank_consent_lifecycle`. Migration 8 creates non-provider-scoped unique indexes, and migration 9 immediately replaces them with provider-scoped indexes. A single final migration would be simpler and avoid carrying an intermediate schema state that no user should ever need.

4: Can you clarify whether `iban` is intended to remain on a local account after unlink/disconnect in [bank_connection_repository](https://github.com/RIP-Comm/sossoldi/blob/a43e822b912206250d870bc1e6f602c0c3ba7757/lib/services/database/repositories/bank_connection_repository.dart#L290-L298)? It is populated from the remote banking account, but `_unlinkAccount` clears the other provider-derived fields and leaves `iban` intact. If `iban` is provider-owned metadata, we should clear it too; if it is local account metadata, keeping it makes sense.

Again, thank you for your time!

